PR #2, opened 2023-05-30
Scantool class and CI

JamesWrigley:
This adds a component class for the scantool.

CC @IvarsKarpics


src/extra/components/scantool.py
```diff
@@ -0,0 +1,166 @@
+import ast
+from warnings import warn
+
+from ..data import SourceData
+
+class Scantool:
+    """Interface for the XFEL scantool (Karabacon).
```
tmichela, line 7:
```suggestion
    """Interface for the EuXFEL scantool (Karabacon).
```

tmichela (approved):
Codewise LGTM, though I haven't used the scantool, so I trust it does what it should.

Maybe I'd turn the attributes into properties, since you nicely documented what they are already, that'd make it easier to look them up from an interractive python session.

src/extra/components/scantool.py
JamesWrigley: Changed to `European XFEL` in ba0e737.

tmichela:
turn the comments in docstring?

JamesWrigley:
Ahhh right so you could call `help()` on them, makes sense.

IvarsKarpics:
LGTM from my (scantool) side.

JamesWrigley:
Replaced attributes with properties in 7b93b0b (and fixed the tests).

JamesWrigley:
Rebased onto `docs`: https://extra--2.org.readthedocs.build/en/2/

PR #3, opened 2023-06-13
Remove import proxy machinery

philsmt:
This unintentionally ended up in the main branch even though it was only an experiment. I don't think such hackery is needed.


JamesWrigley:
Agreed, LGTM!

JamesWrigley (approved):

PR #4, opened 2023-06-13
Initial implementation of Scan component

JamesWrigley:
Leaving this as a draft for now because I still need to write tests, but y'all are welcome to try it on real data to make sure I'm not missing something obvious :sweat_smile: 

Known issues (that may or may not be fixed in the future):
- Automatic resolution detection on non-linear scans is dodgy.
- The first and last detected steps may be part of the motor backlash rather than part of the scan. Haven't thought much about how, but it would be nice to automatically detect and remove them.


src/extra/components/scan.py
```diff
@@ -0,0 +1,293 @@
+from datetime import timedelta
+
+import numpy as np
+from xarray import DataArray
```
philsmt, line 4: `xarray` is really slow to import and can take up to 10s of seconds especially from GPFS environments. Unlike in most other places it does look like this component will always use it, but since it's imported alongside any other component, I would prefer a lazy import.

src/extra/components/scan.py
```diff
@@ -0,0 +1,293 @@
...
+                coordinate with the train IDs (i.e. obtained with
+                [KeyData.xarray()][extra_data.KeyData.xarray]). Raw Numpy arrays are not
+                allowed.
+            resolution (float): The usable resolution of the motor. Ideally this
+                should be slightly above the noise level of the motor (and below the
+                step size of the scan). It will be guessed if not passed explicitly.
+            min_trains (int): The minimum number of trains per-step in the scan. It
+                will be guessed if not passed explicitly.
+        """
+        self._name = "motor"
```
philsmt, line 68: When an `xarray.DataArray` is passed, it may be nice to set this argument manually via an optional argument or try to take it from the `DataArray`'s name (`KeyData.xarray()` will set this to `{source}.{key}` by default).

src/extra/components/scan.py
JamesWrigley: Sure, fixed in 7f69671.

src/extra/components/scan.py
JamesWrigley: Ah nice, didn't know about `DataArray.name`. Added in 7f69671.

JamesWrigley:
I rebased the branch, added some tests, and cleaned up the code a bit. Still not quite happy with it so I'm gonna leave it as a draft for now.

src/extra/components/scan.py
```diff
@@ -0,0 +1,328 @@
...
+
+            self._input_pos = motor.xarray()
+            default_name = f"{motor.source}.{motor.key.removesuffix('.value')}"
+        elif isinstance(motor, DataArray):
+            if not "trainId" in motor.coords:
+                raise ValueError("DataArray of motor positions must have a trainId coordinate")
+
+            self._input_pos = motor
+            default_name = motor.name if motor.name is not None else "motor"
+        elif isinstance(motor, np.ndarray):
```
philsmt, line 91: Amusingly I tried to pass a `DataCollection` and very weird errors appeared 🙃 

To avoid users trying to understand cryptic errors, could you add an `else` branch or generalize the `ndarray` one?

src/extra/components/scan.py
```diff
@@ -0,0 +1,328 @@
...
+        """List of `(position, train_ids)` tuples for each step in the scan."""
+        return self._steps
+
+    @property
+    def positions(self) -> np.ndarray:
+        """Array of positions for each step."""
+        return self._positions
+
+    @property
+    def positions_tids(self) -> list:
```
philsmt, line 120: I would prefer to avoid most abbreviations (_tids_ here) in public APIs, in particular in methods/properties

src/extra/components/scan.py
```diff
@@ -0,0 +1,328 @@
...
+        """List of train IDs for each position."""
+        return self._positions_tids
+
+    def plot(self):
```
philsmt, line 124: Could you please add an optional `axes=None` argument and skip creating a new figure if it's supplied?

philsmt (changes requested):
I gave it a try for a demo in the FXE workshop, and it worked **beautifully**. Doing so I found a few minor usability improvements, but this component is frankly awesome.

src/extra/components/scan.py
JamesWrigley: Fixed in 582f7d8.

src/extra/components/scan.py
JamesWrigley: Renamed to `positions_train_ids` in 582f7d8.

src/extra/components/scan.py
JamesWrigley: Added in 582f7d8.

JamesWrigley:
Rebased and fixed a few things, but there's one more issue with noisy data (noticed in the FXE demo) I want to fix before this is ready for review.

philsmt:
I was also wondering whether the fixtures could directly return the `DataCollection` 🤔 
@takluyver Could there be any state implications by doing so?

JamesWrigley:
Ok, I think this is ready for review now.

Recent changes:
- Added an `intra_step_filtering` parameter for deciding how aggressively to filter trains within a step. Tweaking this fixes the issue I saw with FXE's noisy mono scan.
- Moved the parameter docs into their individual docstrings.
- Made `__repr__()` do something more conventional and recommend `.info()` instead to look at the scan info. Added `.format()` in case someone wants to save the string somewhere.

JamesWrigley:
(if there's no objections I'll probably merge this at the end of next week)

philsmt:
LGTM!

One additional idea I got while remaking the XAS demo on top of `AdqChannel` and `PumpProbePulses`:
* A property yielding tuples of position and `extra_data.by_id` slices.

philsmt (approved):

JamesWrigley:
The problem with using slices is that they'll include any trains that are filtered out within the steps. My inclination is to stick with lists of train IDs, but then again I don't think I've seen any scans that have jumps within the steps :thinking: Lemme think about that.

philsmt:
Oh certainly, but I don't think it's that much of a deal, and slices would be there for convenience anyway. We could have an optional `strict`-style argument that throws an exception if no contiguous slices can be made. I'm curious though, what are the reasons right now a train in the middle can be filtered out?

JamesWrigley:
There's actually none that I can think of, the original reason I did it that way is to be on the safe side in case there was some badly behaving motor (or other device). But I've never seen it happen so it's probably safe to assume that we always want all trains within a step, or make that the default behaviour at least.

JamesWrigley:
I tried to switch to `by_id` by default, but that made the API a bit awkward because it's difficult to convert a `by_id[]` into a list of train IDs to easily find e.g. the length of a step (I know you could check the start/stop values directly but that's less convenient than `len()`) or pass it to `DataArray.select(trainId=...)` :frowning_face: So I stuck with a list of train IDs, but made it include all trains within a step in 90d0603 so we at least get the behaviour of `by_id[]`, which I agree with you should be the default.

PR #5, opened 2023-06-20
PulsePattern component

philsmt:
Finally, this is the first version of the [`PulseTiming`](https://github.com/European-XFEL/EXtra-data/pull/395) successor. I decided to split the original component into multiple pieces for a cleaner interface and add further functionality to each. The entire suite is currently planned to consist of:

* `PulsePattern` to interact with FEL pulses (FEL-only or pump-probe experiments with aligned patterns)
* `PulsePatternPPL` to specifically look for PPL pulses only
* TODO: `PumpProbePattern` to combine FEL and PPL for non-aligned patterns
* TODO `OpticalLaserDelay` combining pulse pattern with BAM corrections and intentional motor-based delay to yield a single delay number for each pulse

The public interface implementation for `PulsePattern` and `PulsePatternPPL` is actually identical and in a shared base class `_PulsePattern`. The reason for this naming is that I wanted to call the version for FEL pulses only `PulsePattern`, since it will cover the majority of use cases for both FEL-only and pump-probe experiments. This PR only contains the core functionality of `PulsePattern` for time reasons, the entire interface is planned to consist of:

* `select_trains(trains)`
* `get_pulse_counts(labelled=True)`: Number of pulses per train as `pd.Series` or `ndarray`
* `get_pulse_ids(labelled=True)`: Pulse IDs from bunch pattern table as `pd.Series` or `ndarray`
* `get_pulse_id_row()`: Fast version of `get_pulse_ids` reading only the first bunch pattern table
* `get_pulse_index(with_pulse_id=True)`: `pd.MultiIndex` with trains and pulses (by ID or number) to label data
* `search_pulse_patterns(with_train_id=True)`: Search contiguous regions of identical pulse pattern with train slices by ID or index
* `trains()`: Iterate over `(train_id, pulse_ids)`
* TODO `label_pulse_data(data, train_ids=None, with_pulse_id=True)`: Attach multi-level index to existing data, say when you loaded pulse-resolved data via `KeyData`

I'm not really happy with the `with_pulse_id`/`with_train_id` flag on some of these function. I think I'd like it to be there so fossils like me can continue using unlabelled `ndarray`, but the name seems clunky. Any ideas?

### DIFF
--- a/src/extra/components/pulses.py
+++ b/src/extra/components/pulses.py
@@ -1,0 +1,481 @@
+"""Interface and utilies to work with pulse-resolved data."""
+
+
+from copy import copy
+from typing import Optional
+import re
+
+import numpy as np
+
+from euxfel_bunch_pattern import PPL_BITS, is_sase, is_laser
+from extra_data import SourceData, KeyData, by_id
+
+
+try:
+    from itertools import pairwise
+except ImportError:
+    # Compatibility for Python < 3.10
+    from itertools import tee
+
+    def pairwise(iterable):
+        # pairwise('ABCDEFG') --> AB BC CD DE EF FG
+        a, b = tee(iterable)
+        next(b, None)
+        return zip(a, b)
+
+
+__all__ = ['XrayPulses', 'OpticalLaserPulses']
+
+
+class PulsePattern:
+    """Abstract interface to pulse patterns.
+
+    This class should not be instantiated directly, but one of its
+    implementationsd `XrayPulses` or `OpticalLaserPulses`. It provides
+    the shared interface to access pulse patterns encoded in the bunch
+    pattern table.
+    """
+
+    # An implementation of this base class is required to implement the
+    # _get_pulse_mask() method.
+
+    # Regular expressions for timeserver control and pipeline data.
+    _timeserver_control_re = re.compile(
+        r'^[A-Z]{3}_(BR|RR)_(UTC|SYS)/TSYS/TIMESERVER$')
+    _timeserver_pipeline_re = re.compile(r'^{}:outputBunchPattern'.format(
+        _timeserver_control_re.pattern[:-1]))
+
+    # Class IDs for timeserver devices.
+    _timeserver_class = 'TimeServer'
+
+    def __init__(self, data, timeserver=None, sase=None):
+        if timeserver is None:
+            timeserver = self._find_timeserver(data)
+
+        self._source = data[timeserver]
+
+        if ':' in timeserver:
+            self._key = self._source['data.bunchPatternTable']
+        else:
+            self._key = self._source['bunchPatternTable']
+
+    @classmethod
+    def _find_timeserver(cls, data):
+        """Try to find a timeserver source."""
+
+        # Try to go by device class first.
+        # By the time the device class was recorded, time servers also
+        # were changed to output the bunch pattern via a pipeline.
+        timeserver_sources = {
+            f'{source}:outputBunchPattern'
+            for source in data.control_sources
+            if (data[source].device_class == cls._timeserver_class and
+                f'{source}:outputBunchPattern' in data.instrument_sources)
+        }
+
+        if len(timeserver_sources) > 1:
+            raise ValueError('multiple timeserver sources found via device '
+                             'class, please pass one explicitly:\n' +
+                             ', '.join(sorted(timeserver_sources)))
+        elif timeserver_sources:
+            return timeserver_sources.pop()
+
+        # Next check for timeserver instrument data.
+        for source in data.instrument_sources:
+            m = cls._timeserver_pipeline_re.match(source)
+            if m is not None:
+                timeserver_sources.add(m[0])
+
+        if len(timeserver_sources) > 1:
+            raise ValueError('multiple timeserver instrument sources found, '
+                             'please pass one explicitly:\n' + ', '.join(
+                                sorted(timeserver_sources)))
+        elif timeserver_sources:
+            return timeserver_sources.pop()
+
+        # Last chance for timeserver control data, likely
+        # empty for more recent data after 2020.
+        for source in data.control_sources:
+            m = cls._timeserver_control_re.match(source)
+            if m is not None:
+                timeserver_sources.add(m[0])
+
+        if len(timeserver_sources) > 1:
+            raise ValueError('multiple timeserver control sources found, '
+                             'please pass one explicitly:\n' + ', '.join(
+                                sorted(timeserver_sources)))
+        elif timeserver_sources:
+            return timeserver_sources.pop()
+
+        raise ValueError('no timeserver found, please pass one explicitly')
+
+    def _make_pulse_index(self, pulse_mask, with_pulse_id=True):
+        """Generate multi index for a given pulse mask."""
+
+        pulse_counts = pulse_mask.sum(axis=1)
+        train_ids = np.repeat(self._key.train_id_coordinates(), pulse_counts)
+
+        if with_pulse_id:
+            train_pulses = [mask.nonzero()[0] for mask in pulse_mask]
+            pulses_label = 'pulseId'
+        else:
+            train_pulses = [np.arange(count) for count in pulse_counts]
+            pulses_label = 'pulseNumber'
+
+        import pandas as pd
+        return pd.MultiIndex.from_arrays(
+            [train_ids, np.concatenate(train_pulses)],
+            names=['trainId', pulses_label])
+
+    def _get_pulse_mask(self, table):
+        """Generate pulse mask from a given bunch pattern table."""
+        raise NotImplementedError('_get_pulse_mask')
+
+    @property
+    def master_clock(self) -> float:
+        """European XFEL timing system master clock in Hz."""
+        return 1.3e9
+
+    @property
+    def bunch_clock_divider(self) -> int:
+        """Divider to generate repetition rate from master clock."""
+        return 288
+
+    @property
+    def bunch_repetition_rate(self) -> float:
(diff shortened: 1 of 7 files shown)


src/extra/components/pulses.py
```diff
@@ -0,0 +1,422 @@
...
+
+        if labelled:
+            import pandas as pd
+            return pd.Series(
+                data=counts, dtype=np.int32,
+                index=self._key.train_id_coordinates())
+        else:
+            return counts
+
+    def get_pulse_id_row(self):
```
JamesWrigley, line 174: Bikeshedding here, what do you think about `get_train_pulse_ids()` instead?

src/extra/components/pulses.py
```diff
@@ -0,0 +1,422 @@
...
+        # pairwise('ABCDEFG') --> AB BC CD DE EF FG
+        a, b = tee(iterable)
+        next(b, None)
+        return zip(a, b)
+
+
+__all__ = ['PulsePattern']
+
+
+class _PulsePattern:
```
JamesWrigley, line 30: I might suggest another property: `is_constant` to detect whether the pulse pattern is constant or not. MID typically uses a constant pulse pattern, but for various reasons the number of pulses in their runs is semi-regularly non-constant, and it could be useful to have a property to check that.

src/extra/components/pulses.py
```diff
@@ -0,0 +1,422 @@
...
+        """Try to identify which PPL seed this data belongs to."""
+
+        instruments = {instrument for src in data.all_sources
+                       if ((instrument := src[:src.find('_')])
+                           in cls._instrument_ppl_seeds.keys())}
+
+        if len(instruments) == 1:
+            return cls._instrument_ppl_seeds[instruments.pop()]
+        elif len(instruments) > 1:
+            raise ValueError('sources from ambiguous instruments {} found,
```
JamesWrigley, line 408: This line is missing a trailing quote:

```suggestion
            raise ValueError('sources from ambiguous instruments {} found, '
```

JamesWrigley (changes requested):
(full disclosure: not an expert with the pulse pattern stuff)

About the `with_pulse_id`/`with_train_id` thing, I can't think of anything particularly clean... Best thing I came up with is something like:
```python
get_pulse_index(slice_by="idx") # or "pulse"
search_pulse_patterns(slice_by="train") # or "idx"
```

I originally thought of `index` instead of `idx`, but that would've felt a bit weird with a function already named `get_pulse_index()`.

JamesWrigley:
BTW, this won't be affected by https://redmine.xfel.eu/issues/151949, right? At least, as far as I can see it doesn't assume that there are 2700 entries in the bunch pattern table.

src/extra/components/pulses.py
philsmt: I'm not sure... how about `peek_pulse_ids`?

src/extra/components/pulses.py
philsmt: Oh, brilliant idea. As it happens, the previous version of the component had that very logic already.

Now I would gravitate towards a method rather than property, is it necessitates a full data read and is thus potentially expensive. Any objection?

src/extra/components/pulses.py
```diff
@@ -0,0 +1,422 @@
...
+
+        if len(sases) == 1:
+            return sases.pop()
+        elif sases == {1, 3}:
+            # SA3 data often contains one or more SA1 sources
+            # from shared upstream components.
+            return 3
+        else:
+            raise ValueError('sources from ambiguous sase branches {} found, '
+                             'please pass explicit sase'.format(
```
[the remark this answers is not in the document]
philsmt: Good catch, it's also capitalized everywhere else. Fixed in f61931c

src/extra/components/pulses.py
```diff
@@ -0,0 +1,422 @@
...
+        timeserver_sources = {
+            f'{source}:outputBunchPattern'
+            for source in data.control_sources
+            if (data[source].device_class == cls._timeserver_class and
+                f'{source}:outputBunchPattern' in data.instrument_sources)
+        }
+
+        if len(timeserver_sources) > 1:
+            raise ValueError('multiple timeserver sources found via device '
+                             'class, please pass explicit one:\n' +
```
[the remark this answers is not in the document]
philsmt: Thanks, that sounds better. Fixed in f61931c

src/extra/components/pulses.py
philsmt: That happens when you add things after running tests :see_no_evil:

tests/test_components_pulses.py
```diff
@@ -0,0 +1,124 @@
...
+    run = MagicMock()
+    run.__getitem__ = lambda self, _: mock_timeserver
+    run.control_sources = {'SPB_RR_SYS/TSYS/TIMESERVER', 'SA1_SOME_DEVICE'}
+    run.instrument_sources = {'SPB_RR_SYS/TSYS/TIMESERVER:outputBunchPattern'}
+    run.all_sources = run.control_sources | run.instrument_sources
+    return run
+
+
+def test_init(mock_data_collection, mock_timeserver, mock_bunch_pattern_table):
+    pp = PulsePattern(mock_data_collection)
```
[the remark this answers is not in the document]
philsmt: True. I'll add a temporary test for now, but as mentioned above after writing the tests I gravitate towards wanting real objects to test with.

philsmt:
> BTW, this won't be affected by https://redmine.xfel.eu/issues/151949, right? At least, as far as I can see it doesn't assume that there are 2700 entries in the bunch pattern table.

Yes, thanks to Python being a sane language we usually never assumed an actual length of the bunch pattern table in our code. It'll *just work*

philsmt:
I have been thinking more about the naming, and concluded that using the generic word `PulsePattern` for FEL-pulses specifically may work out, but gets confusing when comparing to the various implementations including optical lasers. **If** FEL and PPL pulses align it's awesome and obvious, but if not the choice becomes awkward.

So thinking forward to what I want to add next, I would change names to:

* `_PulsePattern` -> `PulsePattern`
* `PulsePattern` -> `FelPulses` (as it is explicitly looking for FEL pulses and nothing else)
* `PulsePatternPPL` -> `OpticalLaserPulses`
* Upcoming class `PumpProbePulses` to combine both also for asymmetric patterns

src/extra/components/pulses.py
JamesWrigley: Yep, that sounds good too :+1: It's mostly the word `row` that threw me because I don't think we use that word anywhere else.

src/extra/components/pulses.py
JamesWrigley: Nah, method should be fine. If reading is expensive then maybe we could also cache the result, like extra-data does in a lot of places?

JamesWrigley: Ehhh nah that sounds fine.

philsmt:
For reference, we decided to remove the picking-the-output-coords arguments for now and re-add them later in a better form. Also rebased now on `master`.

src/extra/components/pulses.py
```diff
@@ -0,0 +1,469 @@
...
+        Yields:
+            (int, ndarray) Train ID and pulse IDs.
+        """
+
+        for train_id, table in self._key.trains():
+            yield train_id, self._get_pulse_mask(table).nonzero()[0]
+
+
+class XrayPulses(PulsePattern):
+    """An interface to X-ray free elctron laser pulses.
```
JamesWrigley, line 313: Minor typo:
```suggestion
    """An interface to X-ray free electron laser pulses.
```

src/extra/components/pulses.py
```diff
@@ -0,0 +1,469 @@
...
+
+    The pulse structure of each train at European XFEL is described by
+    the bunch pattern table and accesssible in recorded data through the
+    timeserver device. This component aids in locating and reading the
+    bunch pattern table, as well as providing utility methods to apply
+    the pulse patterns to other recorded data.
+
+    This class only deals with X-ray pulses of a particular SASE
+    beamline, please see PulsePatternPPL to access pulses of the optical
```
JamesWrigley, line 322: Old name:
```suggestion
    beamline, please see [OpticalLaserPulses][] to access pulses of the optical
```

tests/test_components_pulses.py
```diff
@@ -0,0 +1,213 @@
...
+    assert isinstance(kd2, KeyData)
+    assert kd1.source == kd2.source
+    assert kd1.key == kd2.key
+    assert kd1.train_ids == kd2.train_ids
+
+
+def test_definitions():
+    XrayPulses.master_clock
+    XrayPulses.bunch_clock_divider
+    XrayPulses.bunch_repetition_rate
```
JamesWrigley, line 30: Do these need asserts or something? Not sure what they're testing :thinking:

JamesWrigley (changes requested):

tests/test_components_pulses.py
philsmt: What it does need is instantiation 🙈 

The test was meant as a smoke test those those attributes exist. Their values are in principle arbitrary and may even change, it's to have a central place for reference.

philsmt: Fixed in 5cf7a87

src/extra/components/pulses.py
philsmt: Thanks, fixed in 96a9dd1

src/extra/components/pulses.py
philsmt: Thanks, fixed in 96a9dd1

JamesWrigley (approved):
IANAE, but LGTM!

philsmt:
Thanks for the review!

PR #11, opened 2023-06-23
Add version to package initializer and read during setup

philsmt:
Took the regexp for version string as defined in https://peps.python.org/pep-0440/#appendix-b-parsing-version-strings-with-regular-expressions


takluyver:
LGTM

I guess people accessing `extra.data` or `extra.geom` might end up with some confusion that these are separate, independently versioned packages, but I don't see anything much to do about that.

PR #17, opened 2023-07-11
Final packaging touches

philsmt:
As discussed this changes to a calendar-based numbering scheme, and potentially any future change we need on the way to first release.

@JamesWrigley @takluyver @tmichela

PR #18, opened 2023-07-11
Replace setup.py with pyproject.toml

RobertRosca:
1. Moved to `pyproject.toml` w/ setuptools instead of `setup.py`.
2. Uses setuptools-scm to write a version file to `./src/extra/__version__.py` during install.
3. Changed `src/extra/__init__.py` to import this version file to expose the version information as normal (`extra.__version__`).

I stuck to the default settings for setuptools-scm ([described here](https://github.com/pypa/setuptools_scm#default-versioning-scheme)). In summary this means that:

1. If a tag is checked out, the version will be set to that tag, e.g. if the tag is `2023.2` the version would be `2023.2`.
2. If a branch is checked out and has a distance (commits made) to the latest tag, version is `{next_version}.dev{distance}+{scm letter}{revision hash}`, e.g. I made a tag for this repo and added a commit, now the version is `2023.3.dev1+gd1e34b8`. `dev1` since there is 1 new commit, `gd1e34b8` since the git hash is `gd1e34b8`.
3. If, for some reason, the repo has uncommitted changes then a timestamp gets added to the end.


RobertRosca:
Philipp and I talked about this a bit, summary is that both `pyproject.toml` and `setup.py` would end up being used if/when cython extensions need to be built.

I slightly lean towards keeping everything in `pyproject.toml` and having just `setup(ext_modules=...)` in `setup.py` when required. Philipp was on the side of keeping everything in `setup.py` and having just `[build-system]` in `pyproject.toml`. But we're both pretty neutral since in the end you use both anyway.

---

Copying over stuff from our Zulip:

> As usual things are needlessly complicated. There are actually three versions for the package:
> 
> 1. Import version `extra.__version__`
> 2. Pip version `pip list`
> 3. Conda version `conda list`
> 
> Both the pip and conda versions can come from some fixed string set in their respective files, or they can get the version by reading some like `src/extra/__version__.py`.
> 
> For the nightly install we have a few options:
> 
> 1. Some fixed string like `nightly` as the version.
> 2. Some dynamic version like the current git hash.
> 3. Whatever the latest hardcoded version number in, e.g., `extra/__version__.py`.
> 
> I'm against 3 since it would be pretty confusing to have an actual tagged version number be the current version since that is pretty confusing.
> 
> I'm tied on 1 and 2 since both achieve the goal of showing it's a rolling release. The only real downside of a fixed thing saying `nightly` is that we don't know what version is currently deployed, which would be essential for troubleshooting and checking what's actually installed.
> 
> Having a version number which reflects what's actually installed would mean using something like setuptools-scm to set a 'central' version which is then used by both pip/conda.
> 
> I made a PR here to implement this: https://github.com/European-XFEL/EXtra/pull/18
> 
> This way the version is either the most recent tag, if no changes have been made since that tag, or the most recent tag + some other identifiable information.
> 
> If people are happy with just having the version as some ambiguous `nightly` thing then that doesn't need any changes.

philsmt:
After looking at how `pyproject.toml` and Cython can [co-exist](https://setuptools.pypa.io/en/latest/userguide/ext_modules.html), I'm also fine with keeping metadata in the TOML file. It's kind of more readable this way.

philsmt:
LGTM

takluyver:
LGTM

(I don't personally like the setuptools_scm model, but we don't need to get into that)

JamesWrigley:
Can we remove `scm_letter` from the version? I feel like I'm going to often forget about it and assume the whole token is the commit hash. No biggie if not, otherwise LGTM.

RobertRosca:
> (I don't personally like the setuptools_scm model, but we don't need to get into that)

I also dislike it, this situation (non-tagged installation into shared environments) is pretty much the only one where it makes sense IMO

RobertRosca:
> Can we remove `scm_letter` from the version? I feel like I'm going to forget about it and assume the whole token is the commit hash. No biggie if not, otherwise LGTM.

Looks like functionality is there for this, in the master branch, but a release hasn't been made since last year. For now we'll just have to stick with it and come back to this when a new release gets made of setuptools-scm.

RobertRosca:
Is everybody okay with me creating a tag after merging this?

philsmt:
Could you please summarize again what the versioning scheme will be with this PR?

JamesWrigley:
I think you can edit this tag: https://github.com/European-XFEL/EXtra/releases/tag/1.0.0a0
We should also merge #17 afterwards.

RobertRosca:
> Could you please summarize again what the versioning scheme will be with this PR?

Sure, I did change it a bit with the last commit [`23ae7bc` (#18)](https://github.com/European-XFEL/EXtra/pull/18/commits/23ae7bca6ef41845c71cf7383aa378c90806c33a) and did not explain anything outside the commit message :sweat_smile:

Previously if the most recent tag was `2023.3`, two new commits were added, and the package was installed, then the version would be `2023.4.dev2+{HASH}`. The scheme was `{next_version}.dev{distance}+{scm letter}{revision hash}`. This was a bit weird since the calendar date version tag was being treated as a semver version and just incremented by 1.

Now I set it to use calver instead so the version formatting will be the previous tag if there have not been any changes since the tag was made, or the **current date** + identifying information.

In summary:

- A tag is created, on the master branch e.g. `2023.7.10`
- If no changes have been made since the tag was created then the version will be the tag: `2023.7.10`
- If there are new commits since the tag then the version will be the current date + identifying information
  - So, previous tag is `2023.7.10`, one commit added and package is installed today then the version would be `2023.7.13.0.dev1+g{HASH}` - current date + no. of commits since tag + short hash
  - If another commit is added and the package is installed tomorrow then it would be `2023.7.14.0.dev2+g{HASH}`

So from the version information you can see both when the last automated installation was performed as well as what commit the installation is up to.

JamesWrigley:
>     * If there are new commits since the tag then the version will be the current date + identifying information
>       
>       * So, previous tag is `2023.7.10`, one commit added and package is installed today then the version would be `2023.7.13.0.dev1+g{HASH}` - current date + no. of commits since tag + short hash
>       * If another commit is added and the package is installed tomorrow then it would be `2023.7.14.0.dev2+g{HASH}`

Is the minor version number is overwritten by the date? That sounds pretty confusing to me :thinking: I'm not sure we care that much about 'when' the last deployment was compared to 'what' is deployed, and to me the 'what' is less clear with this scheme. 

To clarify (correct me if I'm wrong @philsmt), the `2023.2.10` versioning that we decided on isn't meant to represent the date of the release, but rather `year.cycle.minor`.

philsmt:
Thanks for clarification @RobertRosca 

> To clarify (correct me if I'm wrong @philsmt), the `2023.2.10` versioning that we decided on isn't meant to represent the date of the release, but rather `year.cycle.minor`.

Exactly that, which is why I was also confused. The first two numbers are solely meant to represent the cycle, and the third one may be any identifier indicating the rolling release. At the least this may be an enumerating integer or the git hash (maybe both?). The actual date is entirely irrelevant. In the interest of brevity it might be nice to also skip any `dev` etc term, so one of:

* `2023.2.4`
* `2023.2+2efa30f0`
* `2023.2.4+2efa30f0`

or similar.

RobertRosca:
I was going to say that it isn't that easy to customise the version past what configuration `setuptools_scm` allows, which isn't much, but with some debugging breakpoints and by pressing "next line" a few thousand times I did find a way (that is not documented by `setuptools_scm` for some reason, even though there are issues/complaints related to this).

It is... mildly hacky, but less hacky than I expected. Basically there's an option to tell setuptools to pull the version number from an attribute in a module:

```toml
[tool.setuptools.dynamic]
version = { attr = "MODULE.ATTRIBUTE" }
```

So we can just define the version scheme you want with some function that calls `setuptools_scm` and then boom it's included as the version.

At build time you are are in the `src` directory. This means you can keep this functionality inside the project, e.g. `src/extra/_version.py` with `{attr = "extra._version.version"}`, but then you end up packaging this stupid file for no reason.

Alternatively, since the cwd is `src`, you can keep the file in the `src` directory and then it won't be included in the package, e.g. `src/get_version.py` with `{attr = "get_version.version"}`.

TBH this isn't really even mildly hacky, it seems pretty reasonable, it was a pain to work out since it isn't documented as an option so I'll make an issue asking to add this functionality to the documentation for setuptools_scm.

philsmt:
One may argue now that having the metadata in `setup.py` again would remove the need for `get_version.py` 😜

RobertRosca:
Well... yes, but so would accepting its default configuration for the version scheme :joy:

JamesWrigley:
What's the status of this now @RobertRosca? :)

philsmt:
As discussed in GangSTAS, LGTM. I will rebase the package rename in https://github.com/European-XFEL/EXtra/pull/17 after that.

PR #20, opened 2023-07-14
Add support to also take pulse data from PulsePatternDecoder devices

philsmt:
While preparing a demo for FXE, I noticed the timeserver is never saved to file at this instrument, there's not even a config group for it. In principle most of the functionality of this component can also be exposed from this information, so I figured it may make sense to offer support for it as well. That has the upside that in the future, users don't even need to care which kind of pulse-pattern-carrying-device is used.

Unfortunately it did make the interface a bit more messy and required a somewhat arbitrary internal interface to abstract between the two, but it also forced me to fix a potential bug in `PulsePattern.peek_pulse_ids()` when the first train actually contains no pulse pattern data.

Marked as draft since it's still missing tests and some more docs.

@JamesWrigley @takluyver

### DIFF
--- a/src/extra/components/pulses.py
+++ b/src/extra/components/pulses.py
@@ -177,23 +177,6 @@ class PulsePattern:
             [train_ids, np.concatenate(train_pulses)],
             names=['trainId', pulses_label])
 
-    def _get_pulse_mask(self):
-        if self._with_timeserver:
-            return self._mask_table(self._key.ndarray())
-        else:
-            mask = np.zeros(
-                (len(self._source.train_ids), self._bunch_pattern_table_len),
-                dtype=bool)
-            node = self._get_ppdecoder_node()
-
-            for i, (pulse_ids, num_pulses) in enumerate(zip(
-                self._source[f'{node}.pulseIds'].ndarray(),
-                self._source[f'{node}.nPulses'].ndarray()
-            )):
-                mask[i, pulse_ids[:num_pulses]] = True
-
-            return mask
-
     def _mask_table(self, table):
         """Mask bunch pattern table."""
         raise NotImplementedError('_mask_table')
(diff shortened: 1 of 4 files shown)


src/extra/components/pulses.py
```diff
@@ -169,7 +310,9 @@ def select_trains(self, trains):
 
         res = copy(self)
         res._source = self._source.select_trains(trains)
-        res._key = self._key.select_trains(trains)
+
+        if res._key is not None:
```
JamesWrigley, line 314: When would `res._key` ever be `None`?

JamesWrigley (changes requested):

src/extra/components/pulses.py
philsmt: Thanks, fixed!

src/extra/components/pulses.py
```diff
@@ -416,64 +553,58 @@ class OpticalLaserPulses(PulsePattern):
     Args:
         data (extra.data.DataCollection): Data to access bunch pattern
             data from.
-        timeserver (str, optional): Source name of a timeserver, only
-            needed if the data includes more than one timeserver or it
-            could not be detected automatically.
+        source (str, optional): Source name of a timeserver or pulse
+            pattern decoder, only needed if the data includes more than
+            such device or it could not be detected automatically.
```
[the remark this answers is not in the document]
philsmt: Thanks, fixed!

src/extra/components/pulses.py
```diff
@@ -339,56 +495,37 @@ class XrayPulses(PulsePattern):
...
```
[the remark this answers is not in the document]
philsmt: Thanks, fixed!

src/extra/components/pulses.py
```diff
@@ -310,16 +456,26 @@ def trains(self):
...
+                yield train_id, pulse_ids[:num_pulses]
 
 
 class XrayPulses(PulsePattern):
     """An interface to X-ray free electron laser pulses.
 
     The pulse structure of each train at European XFEL is described by
     the bunch pattern table and accesssible in recorded data through the
-    timeserver device.
+    timeserver device or in decoded form through pulse patter decoders.
```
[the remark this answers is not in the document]
philsmt: Thanks, fixed!

src/extra/components/pulses.py
philsmt: Got carried away here with "Thanks, fixed!". Proper reply would be, it used to be while I drafted the feature 🙃

src/extra/components/pulses.py
```diff
@@ -36,32 +36,50 @@ class PulsePattern:
...
     """
 
-    # An implementation of this base class is required to implement the
-    # _get_pulse_mask() method.
+    # All methods are built on top of get_pulse_mask and trains()
+    # The default implementations use _mask_table() and _get_ppdecoder_node()
 
-    # Regular expressions for timeserver control and pipeline data.
+    # Number of elements in bunch pattern table.
+    _bunch_pattern_table_len = 3611
```
JamesWrigley, line 43: I'm probably being a little paranoid here, but could the comment link to the docs/redmine ticket where this is stated? :sweat_smile: To me it's a bit of a random number.

src/extra/components/pulses.py
```diff
@@ -127,9 +173,30 @@ def _make_pulse_index(self, pulse_mask, with_pulse_id=True):
...
+                mask[i, pulse_ids[:num_pulses]] = True
+
+            return mask
+
+    def _mask_table(self, table):
+        """Mask bunch pattern table"""
+        raise NotImplementedError('_mask_table')
+
+    def _get_ppdecoder_node(self):
+        """ """
```
JamesWrigley, line 181: Missing docstring?

JamesWrigley (changes requested):
(very minor nitpicking)

src/extra/components/pulses.py
philsmt: Fixed in d7a5c9e. Note that this document is only available internally in Alfresco.

src/extra/components/pulses.py
philsmt: Fixed in be1175d

JamesWrigley (approved):
IANAE, but LGTM!

philsmt:
Thanks for review!
I will monitor performance for the ppdecoder branch and may add a Cython implementation of the hot loop if it escalates for long runs, once we have a Cython infrastructure in place.

PR #25, opened 2023-07-21
Add constant fraction and dynamic leading edge discriminators

philsmt:
Cython implementations of a constant fraction and dynamic leading edge discriminator. These have accompanied me for a long time and saw a lot of fine-tuning over the years, and form a core component of most TOF and REMI analyses. They've been part of various packages in the past to make them semi-available, including `metropc`. I hope this is finally a good place for them to stay.

This PR will need rebasing after the packaging is finalized, and is missing at least some limited form of testing. Primarily I would like to organize all the various code standing in line to get into EXtra.

@takluyver I'd be interested in any comments you have on this code given your Cython experience.

### DIFF
--- a/src/extra/utils/ftd.pyx
+++ b/src/extra/utils/ftd.pyx
@@ -40,7 +40,7 @@ def config_sinc_interpolation(window=None, search_iterations=None):
     bandwidth $1/T$ from a sequence of real numbers $x[n]$:
 
     $$
-    x(t) = \sum_{n=-\infty}^{\infty} x[n] \, {\rm sinc}\left(\frac{t - nT}{T}\right)
+    x(t) = \sum \limits_n x[n] ~ {\\rm sinc} \\frac {t - n T}{T}
     $$
 
     This can be used to find the optimal edge position between samples
@@ -55,14 +55,15 @@ def config_sinc_interpolation(window=None, search_iterations=None):
     somewhat tuned by parameters:
 
     * `window` specifies the number of samples before and after the
-        interpolation points actually used to evaluate $x(t), i.e. the
+        interpolation points actually used to evaluate $x(t)$, i.e. the
         finite boundaries to approximate the infinite sum above. By
         default, up to 200 samples in each direction are used.
 
     * `search_iterations` specifies the number of binary search steps
         taken to find the optimal edge position from interpolated values
         in between two samples. The maximal resolution in samples the
-        interpolation can therefore achieve is $2^-{\rm search_iterations}$.
+        interpolation can therefore achieve with $N$ iterations is
+        $2^{-N}$.
 
     When set, these parameters apply to all discriminator
     implementations and all their use of sinc interpolation.
@@ -89,7 +90,7 @@ def config_sinc_interpolation(window=None, search_iterations=None):
     return dict(window=sinc_window, search_iterations=sinc_search_iterations)
 
 
-cdef inline double _sinc(double x) nogil:
+cdef inline double _sinc(double x) noexcept nogil:
     """Normalized sinc function."""
 
     if x != 0.0:
@@ -98,7 +99,7 @@ cdef inline double _sinc(double x) nogil:
         return 1.0
 
 
-cdef double _sinc_interp(data_t[::contiguous] y_sampled, double x_interp) nogil:
+cdef double _sinc_interp(data_t[::contiguous] y_sampled, double x_interp) noexcept nogil:
     cdef int k, \
         sampling_start = max(<int>floor(x_interp) - sinc_window, 0), \
         sampling_end = min(<int>ceil(x_interp) + sinc_window, y_sampled.shape[0])
@@ -111,15 +112,18 @@ cdef double _sinc_interp(data_t[::contiguous] y_sampled, double x_interp) nogil:
     return y_interp
 
 
-cdef data_t _cfd_sinc_interp(
+cdef int _cfd_sinc_interp(
     data_array_t signal, int i, int j,
-    data_t delay, data_t fraction, data_t walk
-) nogil:
+    data_t delay, data_t fraction, data_t walk, data_t* result
+) except -1 nogil:
+    # Note that the return value of this function is solely used for
+    # exception propagation, the result is returned by pointer.
+
     cdef int int_delay = <int>ceil(delay)
     cdef bint is_integer_delay = int_delay == delay
 
     cdef int k, \
-        sampling_start = max(i - sinc_window, 0), \
+        sampling_start = max(i - sinc_window, int_delay), \
         sampling_end = min(j + sinc_window, signal.shape[0])
 
     # Interpolation is always done on double precision, as single
@@ -129,6 +133,8 @@ cdef data_t _cfd_sinc_interp(
 
     cdef data_t *interp_buf = NULL
 
+    assert i >= int_delay or j >= int_delay, 'indices smaller than delay'
+
     if not is_integer_delay:
         # For non-integer delays, the sinc interpolation includes
         # the same interpolated values in its own summation for every
@@ -136,10 +142,12 @@ cdef data_t _cfd_sinc_interp(
         # a small buffer significantly increases performance.
         interp_buf = <data_t*>malloc(sizeof(delay) * (2 * sinc_window + 1))
 
+        assert interp_buf != NULL, 'unable to allocate interpolation buffer'
+
         for k in range(sampling_start, sampling_end):
             interp_buf[k - sampling_start] = _sinc_interp(signal, k - delay)
 
-    for _ in range(1, sinc_search_iterations+1):
+    for _ in range(sinc_search_iterations):
         middle_pos = (left_pos + right_pos) / 2
 
         middle_value = 0.0
@@ -165,12 +173,12 @@ cdef data_t _cfd_sinc_interp(
     if not is_integer_delay:
         free(interp_buf)
 
-    return <data_t>middle_pos
+    result[0] = <data_t>middle_pos
 
 
 cdef data_t _dled_sinc_interp(
     data_array_t signal, bint negative, int ratio_idx, data_t ratio_value,
-) nogil:
+) noexcept nogil:
     cdef int k, \
         sampling_start = max(ratio_idx - sinc_window, 0), \
         sampling_end = min(ratio_idx + 1 + sinc_window, signal.shape[0])
@@ -184,7 +192,7 @@ cdef data_t _dled_sinc_interp(
         middle_value, middle_pos = 0.0, \
         left_pos = <double>ratio_idx, right_pos = <double>(ratio_idx + 1)
 
-    for _ in range(1, sinc_search_iterations+1):
+    for _ in range(sinc_search_iterations):
         middle_pos = (left_pos + right_pos) / 2
 
         middle_value = 0.0
@@ -204,10 +212,9 @@ cdef data_t _dled_sinc_interp(
 
 def cfd(
     data_array_t signal, data_t threshold, data_t delay,
-    int width=0, data_t fraction=1.0, data_t walk=0.0,
+    data_t width=0.0, data_t fraction=1.0, data_t walk=0.0,
     int interp=EdgeInterpolation.LINEAR,
-    data_array_t edges = None,
-    data_array_t amplitudes = None
+    data_array_t edges = None, data_array_t amplitudes = None
 ):
     """Constant fraction discriminator.
 
@@ -217,18 +224,27 @@ def cfd(
     pulse to peak at its largest value while a negative threshold
     implies the pulse to peak at its smallest value.
 
+    This discriminator can use sinc interpolation both to find the
+    optimal walk crossing as well as to enable real delay values. Please
+    see [extra.utils.ftd.config_sinc_interpolation] for more details.
+
+    Note that enabling both these features at the same time makes the
+    discrimination much more expensive, as repeated nested
+    interpolations are necessary.
+
     Args:
-        signal (array_like): 1D input array with analog signal.
-        threshold (data-type): Trigger threshold, positive values imply
-            a positive pulse slope while negative values correspondingly
-            imply a negative pulse slope.
-        delay (int): Delay between the raw and inverted signal.
+        signal (ArrayLike): 1D input array with analog signal.
+        threshold (np.float32 or np.float64): Trigger threshold,
+            positive values imply a positive pulse slope while negative
+            values correspondingly imply a negative pulse slope.
+        delay (np.float32 or np.float64): Delay between the raw and
+            inverted signal.
         width (int, optional): Minimal distance between found edges,
             none by default.
-        fraction (data-type, optional): Fraction of the inverted signal,
-            1.0 by default.
-        walk (data-type, optional): Point of intersection in the
-            inverted signal, 0.0 by default.
+        fraction (np.float32 or np.float64, optional): Fraction of the
+            inverted signal, 1.0 by default.
+        walk (np.float32 or np.float64, optional): Point of intersection
+            in the inverted signal, 0.0 by default.
         interp (EdgeInterpolation, optional): Interpolation mode to
             locate the edge position, linear by default.
         edges (ArrayLike, optional): 1D output array to hold the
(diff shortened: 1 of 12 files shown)


src/extra/utils/ftd.pyx
```diff
@@ -0,0 +1,354 @@
...
+
+def sinc_interpolate(
+    double[::contiguous] y_sampled, double[::contiguous] x_interp,
+    double[::contiguous] y_interp = None, int window=100
+):
+    # TODO: Useful function that should go somewhere, but doesn't
+    # actually belong to this module.
+
+    if y_interp is None:
+        y_interp = np.zeros(len(x_interp), dtype=np.float64)
```
takluyver, line 38: If `y_interp` is passed, I think you're assuming it's the same size as `x_interp`, right? `i` is used as an index for both.

As you're (sensibly) disabling bounds checking, I would explicitly check this before the loop. It's a small performance cost, but I think functions you can call from Python code should never segfault just because you made a mistake with the arguments.

src/extra/utils/ftd.pyx
```diff
@@ -0,0 +1,354 @@
...
+            y_interp[i] += y_sampled[k] * _sinc(k - x_interp[i])
+
+    return np.asarray(y_interp)
+
+
+cdef int _cfd_fast_pos(
+    data_array_t signal, data_array_t edges, data_array_t amplitudes,
+    data_t threshold, int delay, int width, data_t fraction, data_t walk,
+    int interp
+) nogil:
```
takluyver, line 55: `interp` is unused (same for the matching _neg version below).

Is amplitudes meant to be filled in with something?

takluyver:
As an aside, what does `ftd` stand for?

I can look more next week.

src/extra/utils/ftd.pyx
philsmt: Yes, `interp` and `amplitudes` is a recent addition to the interface I've only added in the DLE implementation yet. As there's plenty of code that expects both functions to have the same interface, I added stubs there. Started on minimal code for `interp` in fa1f645.

src/extra/utils/ftd.pyx
philsmt: Good point, missed that one.

philsmt:
Thanks!

FTD is a semi common abbreviation for fast timing discriminator (hardware). One *could* call the module "peak finding", but these methods are fairly special to find timing pulses with fairly large rise time, i.e. do the same job on digitized signals in software. I've added a docstring to clarify this. There's some documentation missing anyway.

philsmt:
I will fix the CI after https://github.com/European-XFEL/EXtra/pull/18 is merged.

philsmt:
Rebased on the new project structure with `pyproject.toml` and seems to work fine with a `setup.py` purely for building extensions. Note the `package_dir` directive is still necessary.

philsmt:
This PR is now ready for review.

src/extra/utils/ftd.pyx
```diff
@@ -0,0 +1,517 @@
...
+                    raise ValueError('invalid interpolation mode')
+
+                edges[edge_idx] = edge_pos
+                next_edge = i + width
+                edge_idx += 1
+
+                if edge_idx == max_edge:
+                    break
+
+    return np.asarray(edges)[:edge_idx], np.asarray(amplitudes)[:edge_idx], \
```
takluyver, line 333: Is it needed to convert the return values to numpy arrays? They have to be array-like anyway, so presumably `[:n]` works.

philsmt: Both `edges` as well as `edges[:edge_idx]` yield a `MemoryView` object on the Python side. Cython recommends to use the typed memoryview interface for new code rather than the prior numpy integration, which has the unfortunate side effect you need to cast it back to a Python type at the end. On the upside though, any type implementing the buffer protocol is supported as input this way.

philsmt: Ah, and typed memoryview do not require the GIL when passed around 🙂

src/extra/utils/ftd.pyx
```diff
@@ -0,0 +1,517 @@
...
+    if window is not None:
+        sinc_window = <int>window
+
+    if search_iterations is not None:
+        sinc_search_iterations = <int>search_iterations
+
+    return dict(window=sinc_window, search_iterations=sinc_search_iterations)
+
+
+cdef inline double _sinc(double x) nogil:
```
takluyver, line 92:
```suggestion
cdef inline double _sinc(double x) noexcept nogil:
```

From Cython 3.0, functions like this one are considered able to raise exceptions by default, so after each call it will insert a check for an exception.

https://docs.cython.org/en/latest/src/userguide/migrating_to_cy30.html#exception-values-and-noexcept

The same applies to `_sinc_interp`, `_cfd_sinc_interp` and `_dled_sinc_interp`, I think. Naturally if we're using `noexcept` we should be confident these can't raise exceptions.

src/extra/utils/ftd.pyx
```diff
@@ -0,0 +1,517 @@
...
+        middle_pos = 0.0, middle_value
+
+    cdef data_t *interp_buf = NULL
+
+    if not is_integer_delay:
+        # For non-integer delays, the sinc interpolation includes
+        # the same interpolated values in its own summation for every
+        # iteration. Computing these values only once and keep them in
+        # a small buffer significantly increases performance.
+        interp_buf = <data_t*>malloc(sizeof(delay) * (2 * sinc_window + 1))
```
takluyver, line 137: Do we need to check for `malloc` returning NULL?

src/extra/utils/ftd.pyx
```diff
@@ -0,0 +1,517 @@
...
+
+        for k in range(sampling_start, sampling_end):
+            interp_buf[k - sampling_start] = _sinc_interp(signal, k - delay)
```
takluyver, line 140: I *think* an out of bounds write is only prevented here by the values of i & j that we expect from the caller - specifically that they differ by 1. I.e. bad inputs to this function could cause a segfault.

If we know that `j` is always `i + 1`, is it OK to pass only `i` in to the function and calculate `j` within it?

src/extra/utils/ftd.pyx
```diff
@@ -0,0 +1,517 @@
...
+
+    for _ in range(1, sinc_search_iterations+1):
```
takluyver, line 142: If we're not using the loop variable...

```suggestion
    for _ in range(sinc_search_iterations):
```

src/extra/utils/ftd.pyx
```diff
@@ -0,0 +1,517 @@
...
+        middle_pos = (left_pos + right_pos) / 2
+
+        middle_value = 0.0
+
+        if is_integer_delay:
+            for k in range(sampling_start, sampling_end):
+                middle_value += (
+                    (signal[k - int_delay] - fraction * signal[k])
```
takluyver, line 150: I can't see anything that prevents `k - int_delay` being negative, and thus reading out of bounds?

philsmt: Not in this function alone, but `i` is guaranteed in https://github.com/European-XFEL/EXtra/blob/37564eb1df7aad24dab41f5cc7682dd4d39aec2a/src/extra/utils/ftd.pyx#L279 to never be smaller than `int_delay`. As with the case above, generally this function is not meant to be used standalone, and only serves to isolate its stack from the `cfd` function loop.

src/extra/utils/ftd.pyx
philsmt: Doh.

src/extra/utils/ftd.pyx
```diff
@@ -0,0 +1,517 @@
...
+            positions and amplitudes, number of found edges.
+    """
+
+    if edges is None:
+        edges = np.zeros(
+            len(amplitudes) if amplitudes is not None else len(signal) // 100,
+            dtype=np.asarray(signal).dtype)
+
+    if amplitudes is None:
+        amplitudes = np.zeros_like(edges, dtype=np.asarray(signal).dtype)
```
takluyver, line 252: I think the `amplitudes` array is never actually filled.

src/extra/utils/ftd.pyx
philsmt: I think so. Shows how long I have not used `malloc`.

src/extra/utils/ftd.pyx
```diff
@@ -0,0 +1,517 @@
...
+    # positive) or handled specifically in other cases.
+    cdef bint negative = threshold < 0
+    cdef data_t s = 1
+
+    if negative:
+        threshold = -threshold
+        s = -1
+
+    cdef int signal_idx, signal_len = signal.shape[0], \
+        edge_idx = 0, max_edge = min(edges.shape[0], edges.shape[0]), \
```
takluyver, line 393:
```suggestion
        edge_idx = 0, max_edge = min(edges.shape[0], amplitudes.shape[0]), \
```

src/extra/utils/ftd.pyx
philsmt: TIL, thanks!

Due to a change below, I did not add it to `_cfd_sinc_interp`.

src/extra/utils/ftd.pyx
philsmt: Oh joy, I forgot to carry over that part of the implementation. Thanks!

src/extra/utils/ftd.pyx
```diff
@@ -0,0 +1,517 @@
...
+    cdef data_t cur_value, ratio_pos, ratio_value, peak_value = 0.0, \
+        last_ratio_pos = -width
+    cdef bint beyond_threshold = False
+
+    with nogil:
+        for signal_idx in range(signal_len):
+            cur_value = signal[signal_idx]
+
+            if negative:
+                cur_value = -cur_value
```
takluyver, line 408:
```suggestion
            cur_value = s * signal[signal_idx]
```

I would think this is quicker than the extra `if`?

src/extra/utils/ftd.pyx
philsmt: Thanks!

src/extra/utils/ftd.pyx
philsmt: Hmmm, not sure if I did profile it or not. Can give it another try.

src/extra/utils/ftd.pyx
```diff
@@ -0,0 +1,517 @@
...
+    As with the constant fraction discriminator, it is assumed the
+    rising pulse slope points away from zero.
+
+    Args:
+        signal (ArrayLike): 1D input array with analog signal.
+        threshold (data-type): Trigger threshold, positive values imply
+            a positive pulse slope while negative values correspondingly
+            imply a negative pulse slope.
+        ratio_max (data-type, optional): Ratio of leading edge to peak
+            value, 0.6 by default.
```
takluyver, line 371: This should presumably be between 0 & 1? Is it worth saying that in the docstring & maybe even checking it in the function?

src/extra/utils/ftd.pyx
takluyver: Does [numpy.asanyarray()](https://numpy.org/doc/stable/reference/generated/numpy.asanyarray.html) work? That should pass numpy-like arrays through unchanged - but I don't know if it also passes MemoryView through.

src/extra/utils/ftd.pyx
takluyver: Thanks, that makes sense. Even if the function is logically 'part of' its caller, it's nice to be able to verify important properties like this by looking at as small a piece of code as possible.

src/extra/utils/ftd.pyx
```diff
@@ -0,0 +1,532 @@
...
+                elif interp == EdgeInterpolation.LINEAR:
+                    edge_pos = i + (cfd_i - walk) / (cfd_i - cfd_j)
+                elif interp == EdgeInterpolation.SPLINE:
+                    raise NotImplementedError('spline interpolation')
+                elif interp == EdgeInterpolation.SINC:
+                    # As sinc interpolation still makes use of the
+                    # original (possibly negative!) signal, it also
+                    # requires the original walk value for comparions.
+                    edge_pos = _cfd_sinc_interp(signal, i, j, delay, fraction,
+                                                orig_walk)
```
takluyver, line 336: As things stand, I think this line will briefly re-acquire the GIL to check for exceptions.

Normally with a numeric return type, Cython returns `-1` to indicate an exception, and after a call only looks for an exception if that's the return value. This is the same as specifying `except? -1`. But it's not smart enough to do this with your specialising `data_t`, and even adding `except? -1` explicitly produces an error.

We could live with that, as this isn't in a hot loop. If you want to get rid of it, the simplest way might be to 'return' the interpolated value via a pointer, and let the actual return value be a simple int so it can indicate failure.

philsmt: I'm not so sure. Checking the annotated output, the GIL seems to only be acquired when the exception is actually raised. When entering the function, only Cython-internel helper functions are called (`__Pyx_RefNannySetupContext`), presumably to prepare for such an event. There's nothing on the calling site.

In any case, that would be acceptable. I would not consider anything using sinc interpolation to be a high performance path - if you're enabling that, you realize it's going to be slower. To give you an idea, it's a 4x penalty to use it for edge position, and another (!) 20x for real delays.

philsmt: Is it possible Cython 3 changed this to handle the exception check at the called site now instead?

src/extra/utils/ftd.pyx
philsmt: Good idea to clarify, and a C-level check will be as good as free. Funny things would happen if you go negative, and nothing would happen if > 1.

src/extra/utils/ftd.pyx
philsmt: I think this tidbit should apply in this case and you're partially right:

> There is an easy-to-encounter performance pitfall here with nogil functions with an implicit exception specification of except *. This can happen most commonly when the return type is void (but in principle applies to most non-numeric return types). In this case, Cython is forced to re-acquire the GIL briefly after each call to check the exception state. To avoid this overhead, either change the signature to noexcept (if you have determined that it’s suitable to do so), or to returning an int instead to let Cython use the int as an error flag (by default, -1 triggers the exception check).

There still does not seem to be any GIL acquisition, likely because Cython looks at each type of the `data_t` fused type separately and can use it to propagate exception state. But -1.0 is a valid return value of this function, so let's go with your idea of passing by pointer.

src/extra/utils/ftd.pyx
philsmt: Done.

src/extra/utils/ftd.pyx
philsmt: I wanted to go back and add checks, but actually I think none are necessay here:

```cython
    cdef int k, sampling_start = max(i - sinc_window, 0), [...]
```

`sampling_start` is guaranteed to never be smaller than `0`, so the lines above cannot explode irrespective of the values of `i` & `j`. The same is not strictly for the other site with `i - int_delay`, I will add a check.

src/extra/utils/ftd.pyx
philsmt: But when should it ever be an ndarray in that place? On the Cython side, it must always be a memory view of something you cannot determine on the Cython side.

src/extra/utils/ftd.pyx
philsmt: Added validation for public arguments, assertions for internal arguments and found a potential out-of-bounds access.

src/extra/utils/ftd.pyx
philsmt: see above.

src/extra/utils/ftd.pyx
philsmt: You're right, in particular for negative threholds (which are more common).

src/extra/utils/ftd.pyx
takluyver: Right, my mistake. I was thinking in Python terms, where if you pass an array-like object in that's what the function works with. But with the type specification in Cython, it's converted to a memory view whatever you pass in.

takluyver:
LGTM

philsmt:
As discussed on Zulip, I moved the code to `extra.signal` with two Cython modules `_ftd` and `_interpolate` (to house the dangling `sinc_interpolate`).

I will merge once the tests succeed and docs build properly.

PR #29, opened 2023-07-27
Move the __repr__() logic into .info()

JamesWrigley:
This hasn't happened yet, but one could imagine doing something like putting one of these objects into a list or something and __repr__() completely breaking the pretty-printing. Instead we now recommend using .info() like with DataCollection's.


takluyver:
LGTM. I generally try to follow this pattern of having a single line repr and explicit methods to get more detailed info, although I know not everyone agrees.

Another option if people don't like that might be to use the IPython special methods to provide a Markdown/HTML repr, which is not used in a container (unless that's changed recently).

PR #34, opened 2023-08-01
Add get_version.py to excluded package data

RobertRosca:
Did not occur to me that `get_version.py` would be included in global namespace since it is in `src/get_version.py`.

PR excludes the file from being packaged.


philsmt:
Thanks, LGTM

PR #35, opened 2023-08-01
Add high-level interface for ADQ digitizer channels

philsmt:
A high-level component to simplify access to data acquired with fast ADQ digitizers.

This MR contains the most elemental functionality of constructing the component for a single ADQ channel with the usual bells and whistles of EXtra components and access to `raw` data. Future additions will expand this to at least the `zero` algorithm exposing zero-suppressed data, as this will be an elemental part of SQS' data reduction strategy. This functionality is intended to be used **the week after next** (last operation week in 2024/1), so it would be important for me to at least review the API itself and potentially defer implementation changes for later.

The basic idea is to offer specialized loading functions both for the data itself as well as commonly needed processed forms of it. In addition to that, all the processing steps in between these forms are also exposed by public methods operating on data in memory to enable customization.

Data accessors:

* `train_raw()`: Get raw traces by train, essentially the same as direct `KeyData` access with additional preprocessing and sample coordinates. 
* `pulse_raw()`: Get raw traces by pulse. Pulse separation is a major theme in this component, as it is often needed but entirely manual for ADQ digitizers. The implemented method is the result of several years of trying, resting on an empirical offset for the first pulse and the bunch pattern table to slice the trace by train into trace by pulses.
* `train_edges()` Run an edge discriminator for every train. Somewhat similar to the relation between charge-integrating and photon-counting 2D detectors, TOF spectrometers can be operated in analog mode using the raw data as a measure of charge or counting mode having a fast pulse for every particle impact and only taking their position. This method directly loads data and discriminates it into edges by train.
* `pulse_edges()` Same as above, but performs pulse separation in addition.

Processing:

* `preprocessing()` Basic *correction* steps for digitizer data, currently baselevel shift and common mode correction
* `reshape_to_pulses()` Pulse separation on trains
* `find_edges()` Performing edge discrimination

Coming up next, I would like to enclose this component in an `AdqDigitizer` object that allows convenient access to several channels at once.

### DIFF
--- a/src/extra/components/_adq.pyx
+++ b/src/extra/components/_adq.pyx
@@ -1,0 +1,42 @@
+
+# cython: boundscheck=False, wraparound=False, cdivision=True, initializedcheck=False
+
+from cython cimport numeric
+from libc.limits cimport INT_MAX
+from libc.string cimport memcpy
+
+
+def _reshape_flat_pulses(
+    numeric[:, :] traces, numeric[:, :] out, int[:] pulse_ids,
+    int samples_per_pulse, int first_pulse_offset, int clock_ratio
+):
+    """Separate trace by pulse into 1D array."""
+
+    if pulse_ids.shape[0] > out.shape[0]:
+        raise ValueError('pulse output buffer smaller than pulse count')
+
+    cdef int start, end, i, j = -1, \
+        cur_pid, prev_pid = INT_MAX, pid_offset = 0, \
+        num_trains = traces.shape[0], trace_len = traces.shape[1]
+
+    cdef size_t bytes_per_pulse = samples_per_pulse * sizeof(traces[0, 0])
+
+    for i in range(pulse_ids.shape[0]):
+        cur_pid = pulse_ids[i]
+
+        if cur_pid < prev_pid:
+            # Pulse ID decreasing means a new train started.
+            pid_offset = cur_pid
+            j += 1
+
+            if j >= num_trains:
+                raise ValueError('train input buffer smaller than train count')
+
+        prev_pid = cur_pid
+        start = first_pulse_offset + (cur_pid - pid_offset) * clock_ratio
+        end = start + samples_per_pulse
+
+        if end > trace_len:
+            raise ValueError(f'trace too short for pulse at {start}:{end}')
+
+        memcpy(&out[i, 0], &traces[j, start], bytes_per_pulse)
--- a/src/extra/components/adq.py
+++ b/src/extra/components/adq.py
@@ -457,27 +457,6 @@ class AdqRawChannel:
             edges=xr.DataArray(edges, coords=coords),
             amplitudes=xr.DataArray(amplitudes, coords=coords)))
 
-    def _iter_cont_chunks(self):
-        """Iterate over contiguous dataset regions."""
-
-        out_start = 0
-
-        # Calling private EXtra-data APIs, use with caution.
-        for chunk in self._raw_key._data_chunks_nonempty:
-            h5_chunk_len = chunk.dataset.chunks[0]
-
-            for local_start in range(0, chunk.total_count, h5_chunk_len):
-                local_end = min(local_start + h5_chunk_len, chunk.total_count)
-                out_end = out_start + (local_end - local_start)
-                first = int(chunk.first)
-
-                yield chunk.dataset, \
-                    np.s_[first+local_start:first+local_end], \
-                    np.s_[out_start:out_end], \
-                    chunk.train_ids[local_start:local_end]
-
-                out_start = out_end
-
     @property
     def control_source(self):
         """Control source of this digitizer, if found in data."""
@@ -1029,10 +1008,14 @@ class AdqRawChannel:
             self._raw_key.entry_shape, roi)
         out = self._validate_out(out, shape)
 
-        # Loop over contiguous chunks of raw instrument data.
-        for dset, row_sel, out_sel, train_ids in self._iter_cont_chunks():
-            dset.read_direct(out[out_sel], source_sel=(row_sel, *roi))
-            self._preprocess(out[out_sel], out[out_sel])
+        offset = 0
+        for kd in self._raw_key.split_trains(trains_per_part=200):
+            num_trains = int(kd.shape[0])  # Beware of uint64
+            out_sel = out[offset:offset+num_trains]
+            offset += num_trains
+
+            kd.ndarray(roi=roi, out=out_sel)
+            self._preprocess(out_sel, out_sel)
 
         if not labelled:
             return out
@@ -1079,24 +1062,22 @@ class AdqRawChannel:
         # Convert to unlabelled array for use in native code later.
         pulse_ids = pulse_ids.to_numpy()
 
-        # Loop over contiguous chunks of raw instrument data.
-        tmp = None  # Temporary buffer for a single HDF chunk.
-        for dset, row_sel, out_sel, train_ids in self._iter_cont_chunks():
-            pulse_sel = np.s_[pulse_layout.loc[train_ids[0]]['first']:
-                              pulse_layout.loc[train_ids[-1]]['last']]
+        # Drop empty trains for efficient access to train IDs.
+        raw_key = self._raw_key.drop_empty_trains()
 
-            if tmp is None or tmp.shape != dset.shape:
-                # Allocate a new buffer is none yet exists or the
-                # chunk size has changed, which *should* not happen.
-                tmp = np.zeros(dset.chunks, dtype=out.dtype)
+        # Temporary buffer for a single iteration.
+        tmp = np.zeros((200,) + self._raw_key.entry_shape, dtype=out.dtype)
 
-            # Read into the temporary buffer.
-            chunk_out = tmp[:(row_sel.stop - row_sel.start)]
-            dset.read_direct(chunk_out, source_sel=(row_sel, *train_roi))
+        for kd in raw_key.split_trains(trains_per_part=200):
+            pulse_sel = np.s_[pulse_layout.loc[kd.train_ids[0]]['first']:
+                              pulse_layout.loc[kd.train_ids[-1]]['last']]
 
-            self._preprocess(chunk_out, chunk_out)
+            tmp_sel = tmp[:len(kd.train_ids)]
+            kd.ndarray(roi=train_roi, out=tmp_sel)
+
+            self._preprocess(tmp_sel, tmp_sel)
             self._reshape_flat_pulses(
-                chunk_out, out[pulse_sel], pulse_ids[pulse_sel],
+                tmp_sel, out[pulse_sel], pulse_ids[pulse_sel],
                 pulse_layout['length'].max())
 
         if not labelled:
--- a/src/extra/components/adq.py
+++ b/src/extra/components/adq.py
@@ -1,0 +1,1317 @@
+
+from copy import copy
+from inspect import signature, Parameter
+import re
+
+import numpy as np
+import pandas as pd
+
+from extra_data.read_machinery import roi_shape
+from .pulses import XrayPulses
+from .utils import _isinstance_no_import
+from ._adq import _reshape_flat_pulses
+
+
+class AdqRawChannel:
+    """A high-level interface to raw output of ADQ digitizer channels.
+
+    Fast ADQ digitizers are used to acquire analog signals at GHz sample
+    rates. Their onboard FPGA are able to generate different kinds of
+    output from the signal they acquire, like raw data, zero suppressed
+    data or peak integration.
+
+    This component is able to access the raw data saved with these
+    devices in various representations Tand data formats.
+
+    All retrieval methods can perform implicit corrections and other
+    steps like pulse separation, which can be controlled by parameters
+    when initializing this component.
+
+    By default, common mode corrections are applied to compensate for
+    non-uniform baselines as a result of the ADQ's parallel readout
+    architecture. The used sample periodicity is 8 samples, or 16 samples
+    if interleaved. To disable, the component can be initialized with
+    `cm_period` being 0 or less. These corrections also pull the baseline
+    to zero unless specified otherwise with `baselevel`. If common mode
+    corrections are disabled, the signal is still pulled to a `baselevel`
+    if specified. If no baselevel is specified and common mode
+    corrections are disabled, the data remains unchanged.
+
+    Typical configurations for these corrections are therefore:
+
+      * `cm_period > 0 and baselevel is not None`: Common mode
+        correction is performed with baseline pulled to `baselevel`.
+
+      * `cm_period > 0 and baselevel is None` (defualt): Common mode
+        correction is performed with baseline at `0`.
+
+      * `cm_period == 0 and baselevel is not None`: Baseline is pulled
+        to `baselevel`.
+
+      * `cm_period == 0 and baselevel is None`: No modification is done.
+
+    The baseline can be any contiguous slice of the raw data, by default
+    the first 1000 samples are used.
+
+    All processing performed implicitly by the retrieval methods is also
+    available to run explicitly on a set of data.
+
+    Args:
+        data (extra.data.DataCollection): Data to access digitizer from.
+        channel (str): Channel name with or without underscore, e.g.
+            1_A or 3C.
+        digitizer (str, optional): Source name of an ADQ digitizer, only
+            needed if the data includes more than one such device or
+            none could be detected automatically.
+        pulses (extra.components.pulses.PulsePattern): Pulse component
+            to pull pulse information, by default a
+            [XrayPulses][extra.components.XrayPulses] object is
+            constructed unless pulse information are explicitly disabled
+            by passing `False`. Most pulse separation operations require
+            this to be present.
+        interleaved (bool, optional): Whether this digitizer channel was
+            interleaving samples or not, only needed if it could not be
+            detected automatically.
+        clock_ratio (int, optional): Digitizer sampling clock as
+            multiple of the bunch repetition rate (4.5 MHz).
+        sample_dim ({sample, time}, optional): Coordinates for sample
+            dimension if a labelled result is returned, sample by default.
+        first_pulse_offset (int, optional): Sample where the first
+            pulse begins, 10000 by default. This is used to locate the
+            beginning of the first pulse when pulse separation is used.
+        single_pulse_length (int, optional): Samples per pulse for
+            the case all trains only contain a single pulse, when it
+            cannot be inferred from pulse repetition rate.
+        cm_period (int, optional): Apply common mode correction with
+            specified sample periodicity, disabled for non-positive
+            values and by default 8 or 16 when interleaved.
+        baseline (slice or ArrayLike, optional): Contiguous 1D slice
+            of the trace of each train to determine common mode or
+            baseline or direct baseline data to use, :1000 by default.
+        baselevel (number, optional): ADU value to pull the baseline
+            to, None by default. Note that common mode corrections
+            if enabled always pull the baselevel to zero unless
+            specified otherwise here.
+    """
+
+    # 4.5 Mhz, see extra.components.pulses.
+    _bunch_repetition_rate = 1.3e9 / 288
+
+    _adq_pipeline_re = re.compile(r'(\w+)\/ADC\/(\d):network')
+    _adq_channel_re = re.compile(
+        r'digitizers.channel_(\d)_([A-Z]).raw.samples')
+
+    # Set of digitizers using 3G boards with a 392 clock ratio that do
+    # not advertise as such.
+    _3g_digitizer = {'SQS_DIGITIZER_UTC2/ADC/1'}
+
+    def __init__(self, data, channel, digitizer=None, pulses=None,
+                 interleaved=None, clock_ratio=None, sample_dim='sample',
+                 first_pulse_offset=10000, single_pulse_length=25000,
+                 cm_period=None, baselevel=None, baseline=np.s_[:1000]):
+        if digitizer is None or digitizer not in data.instrument_sources:
+            digitizer = self._find_adq_pipeline(data, digitizer or '')
+
+        self._instrument_src = data[digitizer]
+
+        if len(channel) == 2:
+            self._channel_board = int(channel[0])
+            self._channel_letter = channel[1].upper()
+        elif len(channel) == 3:
+            self._channel_board = int(channel[0])
+            self._channel_letter = channel[2].upper()
+        else:
+            raise ValueError('channel expected to be 2 or 3 characters, '
+                             'e.g. 1A or 1_A')
+
+        self._channel_number = ord(self._channel_letter) - ord('A')
+        self._channel_name = f'{self._channel_board}_{self._channel_letter}'
+
+        key = f'digitizers.channel_{self._channel_name}.raw.samples'
+
+        if key not in self._instrument_src:
+            raise ValueError(f'key {key} for channel {channel} not found '
+                             f'in digitizer instrument source {digitizer}')
+
+        self._raw_key = self._instrument_src[key]
+
+        # Try to find control source.
+        device_id = digitizer[:digitizer.index(':')]
+
+        if device_id in data.control_sources:
+            self._control_src = data[device_id]
+        else:
+            self._control_src = None
+
+        if pulses is False:
+            pulses = None  # Explicitly disabled.
+        elif pulses is None:
+            # If none, try to auto-detect XrayPulses for this data.
+            try:
+                pulses = XrayPulses(data)
+            except ValueError:
+                raise ValueError('could not auto-detect pulse information, '
+                                 'please pass explicit pulses object or '
+                                 'disable explicitly with pulses=False')
+
+        self._pulses = pulses
+
+        if interleaved is None:
+            if self._control_src is None:
+                # Cannot be inferred without control source in data.
+                raise ValueError(f'data is missing control source '
+                                 f'{device_id}, please pass explicit '
+                                 f'interleaved flag')
+
+            interleaved = bool(self._control_src.run_value(
+                f'board{self.board}.interleavedMode'))
+
+        self._interleaved = interleaved
+
+        if clock_ratio is None:
+            if device_id in self._3g_digitizer:
+                clock_ratio = 392
+            else:
+                clock_ratio = 440
+
+        self._clock_ratio = clock_ratio * (2 if interleaved else 1)
+
+        if sample_dim not in {'sample', 'time'}:
+            raise ValueError('sample_dim must be one of `samples`, `time`')
+
+        self._sample_dim = sample_dim
+
+        self._first_pulse_offset = first_pulse_offset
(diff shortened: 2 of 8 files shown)


philsmt:
As discussed on Zulip, I moved the code to `extra.signal` with two Cython modules `_ftd` and `_interpolate` (to house the dangling `sinc_interpolate`).

I will merge once the docs build properly.

src/extra/components/_adq.pyx
```diff
@@ -0,0 +1,42 @@
+
+# cython: boundscheck=False, wraparound=False, cdivision=True, initializedcheck=False
+
+from cython cimport numeric
+from libc.limits cimport INT_MAX
+from libc.string cimport memcpy
+
+
+def _reshape_flat_pulses(
```
philsmt, line 9: Unfortunately I no longer find my documentation on this, but I kept this Cython implementation around as it proved significantly faster than a numpy implementation. Note that this method supports a varying number of pulses given the flat shape, while the other reshaping style does not.

src/extra/components/adq.py
```diff
@@ -0,0 +1,1294 @@
...
+    _adq_channel_re = re.compile(
+        r'digitizers.channel_(\d)_([A-Z]).raw.samples')
+
+    # Set of digitizers using 3G boards with a 392 clock ratio that do
+    # not advertise as such.
+    _3g_digitizer = {'SQS_DIGITIZER_UTC2/ADC/1'}
+
+    def __init__(self, data, channel, digitizer=None, pulses=None,
+                 first_pulse_offset=10000, single_pulse_length=25000,
+                 interleaved=None, clock_ratio=None, preprocess_kw={}):
```
philsmt, line 71: I moved `first_pulse_offset`, `single_pulse_length, `preprocess_kw` from individual methods to the initializer to simplify their signature aind since I expect them to a) change rearely and b) only by experts.

src/extra/components/adq.py
```diff
@@ -0,0 +1,1294 @@
...
+
+        self._raw_key = self._instrument_src[key]
+
+        # Try to find control source.
+        device_id = digitizer[:digitizer.index(':')]
+
+        if device_id in data.control_sources:
+            self._control_src = data[device_id]
+        else:
+            self._control_src = None
```
philsmt, line 104: This component will work with only the `INSTRUMENT` data, while the enclosing `AdqDigitizer` will not. There are currently DAQ problems preventing universal acquisition of slow data for ADQ digitizers, but CTRL is working on that.

src/extra/components/adq.py
```diff
@@ -0,0 +1,1294 @@
...
+
+        for offset in range(period):
+            sel = np.s_[offset::period]
+            np.subtract(
+                signal[..., sel],
+                baseline[..., sel].mean(axis=signal.ndim - 1)[..., None],
+                out=out[..., sel], casting='safe')
+
+    @staticmethod
+    def _correct_cm_by_mean(signal, out, period, baseline, baselevel=None):
```
philsmt, line 191: This method is currently not used, but could be exposed in the future. Computing the common mode over multiple trains rather than per train can have advantages in certain situations, but generally I typically only use the train itself (e.g. also in offline calibration for DLDs).

src/extra/components/adq.py
```diff
@@ -0,0 +1,1294 @@
...
+            edges = xr.DataArray(edges, coords=index_or_coords)
+
+            if not with_amplitudes:
+                return edges
+
+            return xr.Dataset(dict(
+                edges=edges,
+                amplitudes=xr.DataArray(amplitudes, coords=index_or_coords)))
+
+    def _iter_h5_chunks(self):
```
philsmt, line 418: This is copied here as it enabled significant performance improvements for `train_raw()` and `pulse_raw`.

src/extra/components/adq.py
```diff
@@ -0,0 +1,1294 @@
...
+                pulse shaped and without amplitudes.
+
+            result (xarray.Dataset): Edge positions and pulse heights,
+                only if train or pulse shaped and with amplitudes.
+        """
+
+        # When given or returning xarray, include digitizer parameters
+        # as attributes
+
+        if _isinstance_no_import(data, 'xarray', 'DataArray'):
```
philsmt, line 849: This is TBD, but seemed less relevant for now.

src/extra/components/adq.py
```diff
@@ -0,0 +1,1294 @@
...
+        elif shape == 'trains':
+            index_or_coords = {'trainId': self._raw_key.train_id_coordinates()}
+
+        else:
+            raise ValueError('shape must be one of `edges`, `trains`')
+
+        return self._shape_edge_result(edges, amplitudes, index_or_coords,
+                                       shape, squeeze_edges, with_amplitudes)
+
+    def pulse_edges(self, sample_dim='sample',
```
philsmt, line 1166: This signature is already the result of as much pruning as seemed sensible, maybe someone has more ideas.

src/extra/components/adq.py
```diff
@@ -0,0 +1,1294 @@
...
+            raise ValueError('sample_dim must be `time` or `sample`')
+
+    def _shape_edge_result(self, edges, amplitudes, index_or_coords,
+                           shape, squeeze_edges, with_amplitudes):
+        """Shape output data of edge finding methods."""
+
+        if shape == 'edges':
+            from . import DelayLineDetector
+
+            edges = DelayLineDetector._build_reduced_pd(
```
philsmt, line 388: Obviously this should move to a generic module, I simply imported it directly for time's sake.

@takluyver  Your comment on splitting up the method really payed off btw.

src/extra/components/adq.py
```diff
@@ -0,0 +1,1294 @@
...
+            shape ({edges, input}, optional): Whether to keep a single
+                edge dimension (`edges`, default) or preserve the input
+                shape (`input`). The latter option returns labelled
+                xarray objects, while the former returns pandas
+                series or dataframe.
+            squeeze_edges (bool, optional): Whether to minimize the edge
+                axis lengh for xarray objects, True by default and
+                ignored when pandas objects are returned.
+            pasha_ctx (pasha.MapContext, optional): Pasha context to run
+                parallelized operations, pasha.ProcessContext by default.
```
takluyver, line 827: I'm always wary of parallel-by-default methods because it doesn't compose well - if you call this from inside code that's already parallelised, you're likely to get a mysterious error about 'daemon children'. And the magic incantation to get past that will be something like `pasha_ctx=pasha.SerialContext()`.

Can I suggest that instead of exposing the implementation, we make the option something like `parallel=N`, and have a special case for 0 or False to mean run in serial?

src/extra/components/adq.py
takluyver: If TBD, maybe throw a NotImplementedError or something?

takluyver (commented):
I haven't managed to get through all of it, but I've done as much as I can for today, and I think I've identified the biggest questions I have, so I'm submitting the review now.

src/extra/components/adq.py
philsmt: Passing an `xarray` still works and computes the correct result, but will replace the coordinates by `dim_N` rather than the original. I would not think that warrants breaking the method completely in this case.

src/extra/components/adq.py
```diff
@@ -0,0 +1,1294 @@
...
+
+        This method may be called with any combination of arguments to
+        determine the number of samples per pulse, with order of
+        precendence in definition order. Calling with no arguments tries
+        to use pulse information the component was initialized with.
+
+        Args:
+            repetition_rate (float): Pulse repetition rate in Hz.
+            pulse_period (int or float): Pulse period in PPT units (int)
+                or seconds (float).
```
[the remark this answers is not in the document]
philsmt: The PPT in its core is 9 MHz (LCM of EuXFEL's 4.5 MHz and REGGAE's 3 MHz), but our timeserver device only ever exposes the 4.5 MHz variant by omitting ever other entry. The only place I know of where one is exposed to 9 MHz is the BAM DOOCS proxy.

But fair enough about the type mixing, I did not revise this code from its original commit.

src/extra/components/adq.py
```diff
@@ -0,0 +1,1294 @@
...
+            elif 'channel_' in k:
+                # Skip other channel's settings.
+                continue
+
+            # Strip board prefix and .value suffix.
+            channel_vals[k[7:-6]] = v
+
+        return channel_vals
+
+    def select_trains(self, trains):
```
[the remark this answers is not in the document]
philsmt: Actually our components generally don't have that so far 🤔 

But I agree it would be useful, albeit sounds like something we could provide generically for anything that has `.select_trains()` already, potentially to be overwritten if more knowledge exists.

src/extra/components/adq.py
```diff
@@ -0,0 +1,1294 @@
...
+                pulse_ids = self._pulses.peek_pulse_ids()
+
+            if pulse_ids is not None:
+                # May either be passed directly or come from pulses
+                # component.
+                try:
+                    pulse_period = set(pulse_ids[1:] - pulse_ids[:-1])
+                except IndexError:
+                    raise ValueError('two or more pulses requires to infer '
+                                     'pulse period') from None
```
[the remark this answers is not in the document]
philsmt: It does, fascinating. This code exists verbatim in multiple other (legacy) places where I'm fairly certain I saw it triggered. Either some behaviour changed, or the line changed after it worked to the current form. Well spotted.

src/extra/components/adq.py
```diff
@@ -0,0 +1,1294 @@
...
+                if omitted.
+            dtype (DtypeLike, optional): Data type for new array if
+                    allocated, by default float32 with common mode
+                    corrections or same type as saved data.
+            cm_period (int, optional): Apply common mode correction with
+                specified sample periodicity if greater than 0, by
+                default 8 or 16 when interleaved.
+            baseline (slice or ArrayLike, optional): Part of the trace
+                of each train to determine common mode or baseline or
+                direct baseline data to use, :1000 by default.
```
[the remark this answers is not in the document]
philsmt: Oh, a 2D shape would indeed break the baseline computation. A stride (on 1D) however would work with incorrect values.

I would fix these concerns with a docstring change though (1D + unstrided). Frankly I'd say if you go against what the dimensionality the method asks for, expect some broadcasting exception. For stride, the method explains what it does, so don't sabotage it.

src/extra/components/adq.py
philsmt: Amusingly it was exactly like that before, a `num_workers` argument that switched to `SerialContext` if `<= 1`. 

Shortly before pushing I thought a `pasha` context would be more powerful. But I did not so much consider the fact you might want to disable it, but rather customize it more. But the disable case is likely stronger.

src/extra/components/adq.py
```diff
@@ -0,0 +1,1294 @@
...
+                             'information') from None
+
+        if not fractional:
+            pulse_period = int(round(pulse_period, 1))
+
+        return self._clock_ratio * pulse_period
+
+    def preprocess(self, data, out=None, dtype=None,
+                   cm_period=8, baseline=np.s_[:1000], baselevel=None):
+        """Preprocess raw digitizer data.
```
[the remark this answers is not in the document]
philsmt: Thanks for the feedback!

I'm certainly not really happy about some of the interface choices, so I hope we can find something better with more eyes. That being said, I have the feeling I may have forgot some context or not stress it enough, so I'll first have to disagree on a few points:

* I understand where you're coming from with detector data, but for digitizers *raw* data indeed has a very particular meaning. ADQ digitizers (and the same names are used for FastADC) have four distinct output modes for the signal they acquire:
  * `raw`: Continuously sampled signal for up to few ms.
  * `zero`: Fixed length slices of the signal as response to a threshold, can be outside of `raw` trigger
  * `apd`: Integrated signal per pulse
  * `peak`: Position of _peaks_ (what constitutes a peak is up to the FPGA here, and geared towards certain use cases)
 
  These may be enabled in any combination and are quite explicit both in device configuration and data structure. The component should expose all of these output modes on equal footing, and while I see the unfortunate collision with *raw*, I feel strongly to use the same terms. That's why I moved away from the first draft still having `.ndarray()` etc, it was exactly a `raw`-focused view I want to remedy. 

* Yes, I am fully convinced that any look on raw data should include common mode correction. As a matter of fact, after contacting the vendor (Teledyne) about my findings I got confirmation it is a known consequence of the parallel readout architecture. I actually hope that this component helps in spreading this correction to any user. That being said, it will not be relevant in every use case depending on the SNR - but when averaging a trace, the common mode *noise* can grow beyond the single trace random noise (!) depending on the board. It made the difference between us seeing ${\rm Xe}^{43+}$ or not.

* I understand the not careful checking of `preprocess_kw` in that typos or wrong values can easily be passed to the initializer before they get checked during actual use. I see the issue in that, my motivation to move them to the initializer was early user feedback I got that they do not want to constantly repeat such settings when loading data. From there I was afraid of having too many arguments to functions, but that's likely preferred to collating them into unchecked dicts. It being a potential no-op is intentional. As mentioned essentially always you should correct for common mode, and in 99% of cases you want to pull some baseline to zero to make signals interpretable.

* I cannot follow how going by *chunk* in the EXtra-data sense does not its current objective of efficiently loading *corrected* data into memory. The common mode correction require a type change to work (even for a baseline pull you *should* go `float`), and the goal is exactly to avoid a full copy of the entire data, which was the case in the very early versions. Going by chunk will at most load one file into memory, which currently stands at 14G for a full 16 channel ADQ412 covering 50s time. A typical run either taken with REMI or XAS will be a couple of minutes, so the goal of this method is to load 50G-200G into memory and it profiled favorably for these cases against loading - casting - correcting. I intentionally did not fully parallelize this method yet, and likely splitting them arbitrarily would be of benefit here.

My own as well as the SQS requirement (covering various TOF and diode timing applications) for this component is to have single-line methods to efficiently obtain *usable* ***raw*** data as well as discriminated edges, not force users to always chain the same 2-3 methods in 99% of cases. This necessitates offering a common mode correction or at least baseline pull and different output layouts as requested. Apart from the method of pulse separation I daresay XAS has the same needs, and even for pulse separation I suspect the same given my EXtra demo some time ago produced identical results.

Based on your feedback and some inspiration from the masked data PR:

* To work around the naming issue, introduce distinct classes from each output mode, i.e. `AdqRawChannel` and `AdqZeroChannel` in the future. These can carry much more generic methods like `.ndarray()`, `.series()`, 
*`.edges()`, though it still leaves by train vs by pulse to be resolved.

* They could be convenient created with an `AdqDigitizer` by something like `AdqDigitizer(run).raw(cm_period=4).ndarray()`, i.e. trying to distribute the configurability via subtypes.

* Expand `preprocess_kw` as suggested in the initializer of output modes that require it and validate it. It will not be needed for other output modes like `AdqPeakChannel`. 

* Likely we can get rid of the word `preprocess` entirely in the arguments, and as suggested split up the methods for manual processing. If you insist we can introduce boolean like `apply_cm`, though I suspect the current design should be understandable given documentation (like I already attempted) and should also only affect very few users.

src/extra/components/adq.py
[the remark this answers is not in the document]
philsmt: I renamed the time-based version to `duration` and kept `period` in PPT units, in line with the pulse components.

src/extra/components/adq.py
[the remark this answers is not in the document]
philsmt: I think I figured it out, it used to be `pulse_ids[1] - pulse_ids[0]` and become the slice in an attempt to make it more general.

src/extra/components/adq.py
JamesWrigley: I'd just like to point out that in my experience vanishingly few people 1) know how, and 2) actually regularly do multiprocessing in Python. I would be in favour of making it parallel by default for the 95% of people who don't do multiprocessing and give the others an informative error message.

philsmt: I went with Thomas' suggestion of `parallel=N`, though using `None` as default to express the `min(10, cpu_cores // 4)` logic. This should make it convenient, but transparent and easy to disable. In hindsight, there's no point (yet?) in doing thread-based parallelization except for `find_edges()`, I will evaluate that later.

src/extra/components/adq.py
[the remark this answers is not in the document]
takluyver: Thanks for the context. :+1: 

> ...the goal is exactly to avoid a full copy of the entire data, which was the case in the very early versions. Going by chunk will at most load one file into memory...

If you have virtual overview files, though, one chunk as seen by EXtra-data can be a contiguous set of data in the virtual overview file, i.e. the entire run. So if you have the RunValidator set up, as we do for FXE, you're back to a full copy of all the data.

> If you insist we can introduce boolean like apply_cm

I don't insist, but maybe in the docstring of the 'raw' methods we could add something like 'to disable this step, create the component with cm_period=0'. Otherwise people have to jump around the docs to figure that out.

How much time do we have for playing with the design? I know you wanted to use this soon.

src/extra/components/adq.py
takluyver: Thanks, that works for me. :slightly_smiling_face: 

In general I think methods that do parallel processing internally should know what kind of parallelism is likely to work, e.g. threads vs processes, and that shouldn't need to be a parameter. Admittedly that's hard if you can pass an arbitrary function in, but then I think passing functions into APIs is usually a kind of awkward design.

philsmt: Yes, I actually toned the arbitrary function part down already a notch. It is using the DLED now by default, which is very robust and should serve pretty much any discrimination needs. I left the arbitrary function part in, as the machinery could easily be used for any other form of peak finding a user might supply. It seemed cruel to force reimplementation just for API design sake.

src/extra/components/adq.py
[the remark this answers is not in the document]
philsmt: > If you have virtual overview files, though, one chunk as seen by EXtra-data can be a contiguous set of data in the virtual overview file

I did not consider that, thanks. My first reaction when reading your comment was something like "a chunk but at most N rows", that should work in all situations then.

> How much time do we have for playing with the design? I know you wanted to use this soon.

I want to deliver something to next week's users, with a particular focus on edge discrimination by pulse. These users would also like to asap see code examples to prepare. Hence I would put emphasis on API design much more than implementation (e.g. the case here about chunking). If you feel we cannot nail down a solid future-forward design without more time, my substitute would be some standalone code installed into our environment for that purpose that we stop supporting at some point (or replace by calls to EXtra).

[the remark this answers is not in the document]
takluyver: To be honest, I think it's already pretty solid, and will be even more so with some tweaks along the lines we've discussed. :+1: 

I think we could come up with something more elegant if we took more time, but I think delivering useful things to users should take precedence over refining APIs, and we're not going to get it perfect however long we take.

src/extra/components/adq.py
```diff
@@ -0,0 +1,1302 @@
...
+            raise ValueError('Common mode must be positive number')
+
+        out = np.zeros_like(data, dtype=np.float32)
+
+        self._correct_cm_by_train(data, out, cm_period, baseline, baselevel)
+
+        return out
+
+    def pull_to_baselevel(self, data, baseline, baselevel):
+        out = np.zeros_like(data, dtype=np.float32)
```
takluyver, line 716: Docstring? If you're making this a public method.

src/extra/components/adq.py
```diff
@@ -0,0 +1,1302 @@
...
+        """Reshape train data to pulse data.
+
+        This method performs pulse separation by splitting the trace
+        acquired by train into individual traces by pulse based on the
+        pulse information the component is initialized with.
+
+        Args:
+            data (ArrayLike): Digitizer trace(s) for one or more trains,
+                last axis is assumed to be sample axis and on a train
+                boundary.
```
takluyver, line 732: Not sure what 'on a train boundary' means here.

src/extra/components/adq.py
```diff
@@ -0,0 +1,1302 @@
...
+
+        # TODO: Support data.ndim > 2
+
+        if not isinstance(data, np.ndarray):
+            data = np.asarray(data)
+
+        if first_pulse_offset is None:
+            first_pulse_offset = self._first_pulse_offset
+
+        pulse_shape = data.shape[:-2] + (
```
takluyver, line 750: Should this be 

```suggestion
        pulse_shape = data.shape[:-1] + (
```

? If you're starting with a 2D array, `[:-2]` gets you an empty tuple.

src/extra/components/adq.py
```diff
@@ -0,0 +1,1318 @@
...
+                names=[f'dim_{i}' for i in range(data.ndim - 1)])
+        else:
+            index = pd.Index(np.arange(data.shape[0]), name='dim_0')
+
+        return self._shape_edges(edges, amplitudes, index)
+
+    def find_edge_array(self, data, labelled=True, squeeze_edges=True,
+                        edge_func=None, max_edges=50, parallel=None,
+                        **edge_kw):
+        """Find signal edges as sparse array.
```
takluyver, line 931: To my mind, a 'sparse array' means a specialised data structure for representing a large array where only a small fraction of values are filled in, using less memory than the normal ('dense') array representation. As in the [scipy.sparse](https://docs.scipy.org/doc/scipy/reference/sparse.html) package, for instance.

From the code, I don't think that's what you mean here? I think you're returning something conceptually like what I'd call a 'ragged' array, i.e. the number of entries along one axis varies, but the entries that are there are always at the start of the axis. I'm not sure I'd use that term either for a regular array with NaNs in, though. :thinking:

src/extra/components/adq.py
```diff
@@ -0,0 +1,1318 @@
...
+        return xr.Dataset(dict(
+            edges=xr.DataArray(edges, coords=coords),
+            amplitudes=xr.DataArray(amplitudes, coords=coords)))
+
+    def _iter_cont_chunks(self):
+        """Iterate over contiguous dataset regions."""
+
+        out_start = 0
+
+        # Calling private EXtra-data APIs, use with caution.
```
takluyver, line 465: Can I again ask that we either switch this to using `.split_trains()`  to make chunks or just do the correction on all the loaded data at once for now?

'Use with caution' is kind of a meaningless comment. The risk with using private APIs is that they change without warning and break the code using them. Caution in writing the downstream code can't prevent that.

src/extra/components/adq.py
```diff
@@ -0,0 +1,1318 @@
...
+
+            Any further keyword arguments are passed to the edge finding
+            method.
+
+        Returns:
+            edges, amplitudes (numpy.ndarray): Tuple of edge positions
+                and amplitudes, only if not labelled.
+
+            result (xarray.Dataset): Edge positions and pulse heights,
+                only
```
takluyver, line 961:
```suggestion
                only if labelled.
```

And likewise for `train_edge_array` and `pulse_edge_array`.

src/extra/components/adq.py
philsmt: It looks like my last push was not yet visible, the docstring is there.

src/extra/components/adq.py
philsmt: Data by train (as opposed to by pulse), but maybe that wording is clearer then.

src/extra/components/adq.py
philsmt: That is intentional. The method is supposed to work for `data.ndim > 2` at some point, and this achieves that already. But I didn't test it yet, hence the `# TODO`. Frankly I do not expect someone trying often, the idea is to support arbitrary sorting on the user side with the last axes being `(..., num_trains, samples)`.

src/extra/components/adq.py
philsmt: TIL, I was not sure how to call them. It is sparse in terms of contained information, but I'll go with ragged.

src/extra/components/adq.py
takluyver: OK, got it. What about "last axis is assumed to be samples within a train"

src/extra/components/adq.py
philsmt: It's there for us to realize it's a private API and should not be used lightly. While it's always better to rest on public APIs, I do want to be able to use private APIs in EXtra-data. If we change it, we have to change it in EXtra as well. We should always have tests to tell us that.

I left this part in to save time. I fully agree this is not robust for virtual overview files and needs to change, but they are not used at SQS, and there is little operational pressure to change implementation details.

src/extra/components/adq.py
takluyver: I see, but in that case, I think there's something else wrong. `pulse_ids` is documented as 'Pulse IDs of a single train', but `len(pulse_ids)` is used for the output shape. Shouldn't it be `len(pulse_ids) * n_trains` ?

philsmt:
Please note I decided for the name `AdqRawChannel` as it seemed more forward-compatible with any decisions we take. If we go for a single big class for all output modes, it's easier to drag it around as an alias to `AdqChannel`, while the name's already there if we keep them separate.

src/extra/components/adq.py
philsmt: Oops, thanks.

src/extra/components/adq.py
philsmt: Yes, the docstring is wrong due to copypasta. Well spotted. It must be all pulse IDs, as it explicitly supported different number of pulses.

philsmt: But actually, I intended to use the internal pulse information, so there's also a mistake in the implementation. This can be relaxed later, but for now the methods only support the data the component also returns.

takluyver: Thanks, that makes sense now.

src/extra/components/adq.py
philsmt: Yes, that is probably easier to understand than my attempt at "concatenated pulse traces".

src/extra/components/adq.py
philsmt: It's fine, I meant to do some profiling first to find an equally efficient solution, but any likely performance loss is acceptable now. It looks like I can even maintain the cast-on-copy behaviour, which was the original reason I went this route.

philsmt: Please have a look. For now I simply hardcoded `trains_per_part=200`, which for average trace lengths should yield ~500 MiB data per iteration. This could be done smarter in the future.

src/extra/components/adq.py
```diff
@@ -0,0 +1,1302 @@
...
+                          fractional=False):
+        """Get number of samples per pulse.
+
+        Digitizer traces are acquired by train, by may contain the data
+        of several pulses, with each pulse covering a certain number of
+        samples depending on the digitizer and machine repetition rates.
+
+        This method may be called with any combination of arguments to
+        determine the number of samples per pulse, with order of
+        precendence in definition order. Calling with no arguments tries
```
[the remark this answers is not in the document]
philsmt: The way the logic is implemented made it somewhat convoluted to enforce it, but I reworded the docstring to indicate the first non-`None` argument is used, or internal pulse information else.

src/extra/components/adq.py
takluyver: Thanks! I'm happy with that.

takluyver:
LGTM!

philsmt:
Thank you for review!

PR #37, opened 2023-08-03
Add an installation script and installation docs

JamesWrigley:
I sneaked in a minor change to the docs (1ad0135), but the meat of the PR is adding a shell script to handle updating the version of extra deployed in our current environment. It will:
- Do nothing if the latest version is already installed.
- Throw an error if a newer version is available but requires additional dependencies to be installed.
- Install the latest version with `--no-deps` as long as no dependencies need to be installed.

If you run the script now, you'll get:
```
[xsoft@max-exfl461 ~]$ ./install-extra.sh
 - EXFEL modulepath enabled

Loading exfel-python/202302
  Loading requirement: mambaforge/22.11
Nothing to do, installed version is already up-to-date: euxfel-EXtra @ git+https://github.com/European-XFEL/EXtra.git@505f355e4aa1b3a5e8916642ca10da160378d953
```

@RobertRosca, I was also thinking that the cron job could use this instead of running `pip` blindly?


JamesWrigley:
(if there's no objections I'll probably merge this at the end of next week)

install-extra.sh
```diff
@@ -0,0 +1,56 @@
+###########################################################################################
+#                                                                                         #
+# Note: if this script is updated make sure to update the copy in xsoft's home directory! #
+#                                                                                         #
+###########################################################################################
+
```
RobertRosca, line 6:
```suggestion

set -euxo pipefail

```

For this kind of thing failing fast seems like the best option. There are a few places where something could fail but the script might carry on and do unintended things, like if the module command fails.

RobertRosca (changes requested):
Oops left this as a pending review with one joke comment... went back and added a serious one to make up for that :stuck_out_tongue:

README.md
```diff
@@ -20,3 +20,23 @@ rebuild the docs while editing, run:
...
+
+If you want to update it manually, you must follow these steps:
+1. SSH to Maxwell as the `xsoft` user
+1. Run `./install-extra.sh`
+
+Note: make sure to update the copy in `xsoft`'s home directory if the script is
+updated:
+```bash
+rsync -a --progress install-extra.sh xsoft@max-exfl-display.desy.de:/home/xsoft
+```
```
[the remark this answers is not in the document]
JamesWrigley: Umm, I'm not sure I feel comfortable with that one :see_no_evil:

install-extra.sh
JamesWrigley: Good point, added (and tested) in 12f2ea9.

JamesWrigley:
Moved `install-extra.sh` to `docs/` in 9318cdd (discussed on Zulip).

RobertRosca:
LGTM! (also on Zulip)

RobertRosca (commented):
More official LGTM

JamesWrigley:
Oops sorry, that was a bit premature of me :sweat_smile:

PR #38, opened 2023-08-04
Add extra.utils.imshow2 for easier image plotting

JamesWrigley:
I wrote this function ages ago and I'm constantly copying it over to new notebooks, so I figured it might be useful to put here.

### DIFF
--- a/src/extra/utils/misc.py
+++ b/src/extra/utils/misc.py
@@ -30,3 +30,57 @@ def find_nearest_value(array, value: Any) -> Any:
     """
 
     return array[find_nearest_index(array, value)]
+
+
+def imshow2(image, *args, lognorm=False, ax=None, **kwargs):
+    """Display an image with reasonable defaults.
+
+    This function wraps [plt.imshow()][matplotlib.axes.Axes.imshow] to
+    automatically set some defaults:
+
+    - Try to set `vmin`/`vmax` to reasonable values. Note that setting
+      `vmin`/`vmax` is incompatible with the `norm` argument, so they will only
+      be set if `norm` is not passed.
+    - Use an `auto` aspect ratio if the images aspect ratio is too skewed
+      (useful for displaying heatmaps).
+    - Set `interpolation="none"`.
+
+    All arguments other than the ones listed below are passed to
+    [plt.imshow()][matplotlib.axes.Axes.imshow], and explicitly passing any of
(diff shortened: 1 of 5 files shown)


src/extra/utils/misc.py
```diff
@@ -30,3 +30,54 @@ def find_nearest_value(array, value: Any) -> Any:
...
+            default to [plt][matplotlib.pyplot] if none is explicitly passed.
+
+    """
+    if ax is None:
+        # Try to avoid repeatedly importing matplotlib by looking in sys.modules
+        # first.
+        if "matplotlib.pyplot" in sys.modules:
+            plt = sys.modules["matplotlib.pyplot"]
+        else:
+            import matplotlib.pyplot as plt
```
takluyver, line 63:
```suggestion
        import matplotlib.pyplot as plt
```

Importing something a second time is cheap - it's basically just doing the same lookup in `sys.modules` that you do here. So there's no need to complicate it. :slightly_smiling_face: 

```
In [1]: %time import matplotlib
CPU times: user 257 ms, sys: 235 ms, total: 492 ms
Wall time: 445 ms

In [2]: %time import matplotlib
CPU times: user 21 µs, sys: 0 ns, total: 21 µs
Wall time: 32.2 µs
```

src/extra/utils/misc.py
```diff
@@ -30,3 +30,54 @@ def find_nearest_value(array, value: Any) -> Any:
...
+
+        ax = plt
+
+    imshow_kwargs = dict(interpolation="none")
```
takluyver, line 67: Should the docstring mention this too?

src/extra/utils/misc.py
JamesWrigley: Ah cool, I thought it would re-import the whole thing. Fixed in 7906c54.

philsmt (approved):
Oh yes, this looks like it may save some time regularly. Not a fan of the name, but I also don't got any other ideas.

Nothing to complain, just an idea: I regularly use `norm=LogNorm()`  and always forget to add `from matplotlib.colors import LogNorm`. Maybe extend the interface by some `lognorm=True` option?

Still LGTM.

JamesWrigley:
> Nothing to complain, just an idea: I regularly use `norm=LogNorm()` and always forget to add `from matplotlib.colors import LogNorm`. Maybe extend the interface by some `lognorm=True` option?

Good idea, added in b66a3e8 :+1:

PR #39, opened 2023-08-04
Update the changelog

JamesWrigley:
- Replace links to API sections with links to the classes themselves.
- Add an 'Unreleased' section to the changelog.

Maybe we could try updating the changelog in any PRs that are noteworthy, instead of going through commits before a release? I think that would be particularly useful given extra's rolling release model.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -2,15 +2,32 @@
 
 # Changelog
 
+## Unreleased
+
+!!! note
+    All of the changes here are deployed to our current environment, even though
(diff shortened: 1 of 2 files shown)


philsmt (approved):

philsmt:
Should every PR then include an addition to the changelog?

JamesWrigley:
Every PR that has some noteworthy change, yeah. e.g. I probably wouldn't add an entry for #37 but definitely for #38.

PR #41, opened 2023-08-08
Add tag/diff links to the changelog

JamesWrigley:
This makes it easier to see exactly what changes occurred in between releases. Also added a link to #40 in its changelog entry (I missed that before, sorry :see_no_evil:).


philsmt (approved):
Right, sorry! LGTM this.

PR #45, opened 2023-08-12
Activate env modules in install script

RobertRosca:
Adds `source /usr/share/Modules/init/bash` to the script.

Required since module commands not available in cron shell environment.


JamesWrigley:
LGTM!

JamesWrigley (approved):
(more official LGTM! :stuck_out_tongue: )

PR #49, opened 2023-08-24
Reorganize the component docs

JamesWrigley:
The components page is getting... long :sweat_smile: So I figured it made sense to split them up a bit. I'm thinking that not every future component needs it's own sub-section and they can still be documented in `components/index.md`, just the ones that are related to each other. I also added a list of components to make them easier to find, I'm not sure how but it would be nice to automate that in the future.


philsmt:
Seems like a good idea, thanks. LGTM!

philsmt (approved):

PR #51, opened 2023-08-26
Minor fixes for the Scan component

JamesWrigley:
I noticed a couple of things while using this at MID:
- The motor positions were previously set to the position of the first train in the step, which is a bit unexpected. Fixed in bda7451.
- Handling motor backlash manually is annoying. Added the ability to remove it automatically in 8db2108.


JamesWrigley:
(merging because it's very annoying to work around :see_no_evil: )

Issue #23, opened 2023-07-18
Run manual CI instantiating components across real proposal data


philsmt:
There is lots of data out there which may break all kind of assumptions we inevitably have to make when writing components. Similar to offline calibration where we have manual CIs to test against real detector data, we could have such a CI that tries to instantiate various general purpose components across a lot (all?) runs present on GPFS or dCache and the results, e.g. exceptions occuring.

PR #53, opened 2023-08-29
Add an XGM component

JamesWrigley:
I've wanted this for a long time :grinning: 

Some things to consider:
- Should we use `pint` or something for the photon energy/wavelength properties, or indicate the units in the name? (e.g. `photon_energy_kev`)
- I've only added support for what I think are the most common/useful properties used, but if there are any that you think are missing feel free to mention them and I'll add a method or something for them.
- The XGM device uses the word 'bunches' in the device properties, but I opted to go with 'pulses' in this class instead for consistency with other components (e.g. `XrayPulses`).
- I personally haven't used the XGMs from SASE 1/3 much, but I added the `default_sase` argument to make working with those easier. Not sure if that's a good interface or not, I'd be interested to hear what others who use those XGMs think.


docs/components/index.md
```diff
@@ -3,10 +3,17 @@
 This module contains classes that abstract various Karabo devices to make access
 easier.
 
+Component index:
+
+- [XGM][extra.components.XGM]
```
tmichela, line 8: Can we sort components alphabetically? 🙈

src/extra/components/xgm.py
```diff
@@ -0,0 +1,353 @@
...
+    ```
+    """
+
+    # There are three different XGM device classes, though all are very
+    # similar. DoocsXGM is the main one, and DoocsXGMD and DoocsXGMReduced are
+    # used by FXE and SQS respectively. For more information see the original MR
+    # to create them: https://git.xfel.eu/karaboDevices/doocsDevices/-/merge_requests/117
+    DEVICE_CLASSES = ["DoocsXGM", "DoocsXGMD", "DoocsXGMReduced"]
+
+    def __init__(self, run, device=None, default_sase=None):
```
tmichela, line 47: At HED it is typical to always record 2 XGM, one right after the ondulators and one just before the experimental hutch. (I would assume all instrument do something similar).

I think it may be useful to have an easy filtering in the device name, so that one could do: `XGM(run, "SA2")`/`XGM(run, "HED")` or `XGM(run, "XTD1")`/`XGM(run, "XTD6")` without having to type the complete device name or rely on an alias.

src/extra/components/xgm.py
```diff
@@ -0,0 +1,353 @@
...
+        self._intensity = { }
+        self._npulses = { }
+        self._mean_pulses = { }
+        self._max_pulses = { }
+        self._photon_flux = None
+
+        self._extra_sases = [sase for sase in [1, 3]
+                             if f"pulseEnergy.numberOfSa{sase}BunchesActual" in self._control_source]
+
+        from scipy.constants import h, c, e
```
tmichela, line 112: any reason not to have this at the top of the file?

src/extra/components/xgm.py
```diff
@@ -0,0 +1,353 @@
...
+        self._wavelength = self._control_source["pulseEnergy.wavelengthUsed"].as_single_value()
```
tmichela, line 113: I assume the impact of reading data here is, in most cases, painless (I did not try it) but I think I prefer avoiding any data reading in object initialization.
What do you think about moving this in the `wavelength` propertly when it's called (turning it to a `cached_property` for e.g.)

src/extra/components/xgm.py
```diff
@@ -0,0 +1,353 @@
...
+        self._photon_energy = (h * c / e) / (self._wavelength * 1e-6)
+
+        self._doocs_server = self._control_source.run_values()["location.value"]
```
tmichela, line 116: I would also move this out of the init, as this would fail if we try to instantiate the object on an Unioned DataCollection, and it looks like it's not used anywhere else that the `doocs_server` property.

Now that I think about it, it may be worth adding a test with combined runs for this component :p

src/extra/components/xgm.py
```diff
@@ -0,0 +1,353 @@
...
+            if pg == PropertyGroup.MAIN:
+                key = "pulseEnergy.numberOfBunchesActual"
+            else:
+                key = f"pulseEnergy.numberOfSa{pg.value}BunchesActual"
+            self._npulses[pg] = self._control_source[key].xarray()
+
+        return self._npulses[pg]
+
+    def mean_pulses(self, sase=None) -> float:
+        """The mean number of pulses in a run.
```
tmichela, line 210:
```suggestion
        """The mean number of pulses.
```

Technically `DataCollection` is more correct than `run`, although maybe `run` is clearer for people? 🤔 
Maybe removing it entirely is best since it's implicit it applies to the data we've instantiated the class with?

opinions?

note: this would apply to a few other places.

tests/conftest.py
```diff
@@ -1,17 +1,58 @@
...
 from extra_data.tests.mockdata import write_file
-from extra_data.tests.mockdata.xgm import XGM
 from extra_data.tests.mockdata.motor import Motor
+from extra_data.tests.mockdata.base import DeviceBase
 
 from .mockdata.timeserver import Timeserver, PulsePatternDecoder
 
 
+# This is a more accurate representation of an XGM than the XGM class from the
+# extra_data tests (it has some incorrect property names).
```
tmichela, line 15: not for this PR, but does anything prevent us from fixin the extra_data mock class?

JamesWrigley:
While fixing those things another thought crossed my mind: do we want to automatically set `default_sase` based on the topic the XGM is in? e.g. if we're looking at SQS data and we select `SQS_DIAG1_XGMD/XGM/DOOCS`, should `default_sase` be set to 3? Whereas if we select `SA1_XTD2_XGM/XGM/DOOCS` it's set to 1?

tmichela:
I couldn't tell, but probably @philsmt has a feel for that?

philsmt:
The downstream XGMs (definitely SQS, but FXE and SPB look the same) in SA1/3 acquire data in the same way as the tunnel XGMs and have those extra datasets, but here it's entirely useless to ever look at the other SASE as the those pulses cannot even enter those sections (a SA1 pulse will never reach the SQS XGM in their hutch, you measure only noise). So yes, you want to stick even more to the default SASE.

JamesWrigley:
Just to confirm, if an SQS run has a SASE 1 and SQS XGM and a user selects the SASE 1 XGM, you'd want to stick to the 'run default' of SASE 3? Instead of going by the topic the XGM is in and select SASE 1?

docs/components/index.md
JamesWrigley: To a degree yes :stuck_out_tongue: My thinking was that the component index should display the top-level components (i.e. not in a subsection like `Scans` or `Pulse patterns`) first, and then components in subsections. So I'd say that the top-level components and the ones in subsections can be ordered alphabetically, but the ordering should generally match what's displayed in the navigation bar on the left.

src/extra/components/xgm.py
JamesWrigley: Added in a8e7c50, now it accepts case-insensitive unique substrings. I also added support for recognizing alias names for convenience.

src/extra/components/xgm.py
JamesWrigley: Yeah scipy can take a couple seconds to import on Maxwell, I put it in here so it doesn't slow down importing other components.

src/extra/components/xgm.py
JamesWrigley: Sure :+1: Fixed in a8e7c50. I also made some things that were previously properties into methods so people can expect them to possibly take a bit longer to read the data.

src/extra/components/xgm.py
JamesWrigley: I've removed `run` from these docstrings, but I kept it for the constructor because I think it's clearer. I don't feel very strongly about it though.

philsmt:
Not quite, I was referring to looking at the SA1 dataset in the SQS XGM, i.e. looking at an XGM native to an instrument. In the case you describe, i.e. looking at an XGM not native to your instrument, I would raise an exception if no explicit `default_sase` is passed.

The `XTD2` XGM (i.e. the one behind the SA1 undulator) will show the pulse energy of intended SA1 bunches and whatever 
the bunches intended for SA3 may have generated (generally nothing). The `XTD10` XGM on the other hand shows the pulse energy the already used SA1 bunches generate in the SA3 undulator and the pulse energy of actual SA3 bunches. For the latter there's a clear default of looking at the `XTD10` SA3 pulses, but from an SA3 perspective there's no obvious default for the `XTD2 XGM.

JamesWrigley:
Sweet thanks, I'll change it so that it'll throw an exception if an XGM from a different topic is selected :+1:

JamesWrigley:
Fixed in 33b5e8f.

JamesWrigley:
Updated the plots in 88fda6f to include run metadata.

JamesWrigley:
Changes in 1d54185:
- Added errors to the train/pulse plots
- Added `XGM.info()`
- Improved some docstrings

src/extra/components/xgm.py
```diff
@@ -0,0 +1,498 @@
...
+          Mean pulses:           {self.mean_pulses():.3f}
+
+          DOOCS server:          {doocs_server}
+        """
+
+        print(dedent(info_str))
+
+    def __repr__(self):
+        kev = self.photon_energy()
+        return f"<XGM for {self.control_source.source} with nominal photon energy {kev:.1f} KeV>"
```
takluyver, line 498:
```suggestion
        return f"<XGM for {self.control_source.source} with nominal photon energy {kev:.1f} keV>"
```

Picky, I know, but the kilo prefix is small k.

src/extra/components/xgm.py
```diff
@@ -0,0 +1,498 @@
...
+            |xgm = XGM(run, run.alias["spb-xgm"]["data.intensityTD"]) |
+            |                                                         |
+            |# Or alias name                                          |
+            |xgm = XGM(run, "spb-xgm")                                |
+            |                                                         |
+            |# Or a unique (case-insensitive) substring among XGMs    |
+            |xgm = XGM(run, "spb")                                    |
+            -----------------------------------------------------------
+    Out[2]: <XGM for SPB_XTD9_XGM/XGM/DOOCS at 9.3 KeV>
+    ```
```
takluyver, line 107: I see we've done this in the scan components as well, but I don't particularly like the ASCII art Jupyter notebook styling, either in the source or in the rendered docs. :shrug:

JamesWrigley: I see, why is that? For context, the reason I made those over e.g. a Python interpreter output is because I felt that it would be familiar to more people, since most people will interact with this thing from a notebook.

takluyver: I haven't really got reasoning here, I just think it looks rather ugly, and even if people are most used to notebooks, I'd prefer to have examples presented without the ASCII art. It's not important, so if you prefer it enough, feel free to stick with it. I'm just registering that I'm not sold on it. :wink:

src/extra/components/xgm.py
JamesWrigley: Oops :facepalm: Fixed in df3e138.

src/extra/components/xgm.py
JamesWrigley: But I love ASCII art :heart_eyes: I dunno, I don't feel that strongly about it but I do think it's a better example than an interpreter session so I'm inclined to stick with it.

(I spent a bit of time trying to copy the HTML from a Jupyter notebook cell into the markdown docstring, but that didn't go very well...)

src/extra/components/xgm.py
```diff
@@ -0,0 +1,498 @@
...
+            sase (int): Same meaning as in
+                [XGM.intensity()][extra.components.XGM.intensity].
+        """
+        pg = self._check_sase_arg(sase)
+        if pg not in self._max_pulses:
+            self._max_pulses[pg] = int(self.npulses(sase).max().item())
+
+        return self._max_pulses[pg]
+
+    def npulses_is_constant(self, sase=None) -> bool:
```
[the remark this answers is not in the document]
JamesWrigley: Yeah I think you're right, I replaced `mean_pulses()` with `nominal_pulses()` in df3e138. Not entirely sold on the naming, another option would be to have e.g. `npulses(all_trains=False)` but I'm not sure I like that API much better.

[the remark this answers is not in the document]
takluyver: I think I agree with you on having a separate method rather than a parameter, but I'm also not sure about the naming.

Maybe `.npulses()` should be the one giving you a single number (I'm imagining that works for >90% of experiments, but that guess might be wrong) and have a longer name like `npulses_by_train()` for getting an array?

The closest parallel I can think of with what we already have is in the multi-module detector components. They have a method `.frames_per_train()` to get a single number if possible, and a pre-computed pandas series as `.frame_counts` for the count in each train. But that's different enough that I don't think we need to copy it.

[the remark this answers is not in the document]
JamesWrigley: That sounds good to me, renamed in 3a40ea3.

JamesWrigley:
New changes:
- Added a Pint registry in fdb42bf. It caches to a user-local directory to improve load times (from 750ms to 20ms in my measurements on Maxwell), and has an extra `xfel` context with a transformation from length to energy so we can convert wavelength to keV. The intention is for that registry to be used by any other things in EXtra that need units.
- Added units to the XGM component in f61e896:
  - `XGM.wavelength()` and `XGM.photon_energy()` now return `Quantity`'s directly. It is a possibility that someone will ignore the fact that these return unitful values and do e.g. `xgm.wavelength() * 10` to get something in angstroms instead of `xgm.wavelength().to("angstrom")`, in which case they'd end up with the wrong units (could be a problem if they pass it to a function using those units). But I think this is an acceptable risk because:
    - This is a new component so there's no risk of backwards incompatibility.
    - The documentation explicitly warns against this (one would hope that people would read the docs at least to find units since it's not in the function name).
    - It does make conversions very nice:
      image
  - `XGM.intensity()` also has units through [pint-xarray](https://pint-xarray.readthedocs.io/en/latest/index.html). I added it for consistency and because it's nice to have the units displayed when printing the `DataArray`, but I have a hard time coming up with a situation where it's actually useful (personally I only use the pulse-resolved values as an I0) and pint-xarray says it's still an experimental package. So I don't feel very strongly about including it.

JamesWrigley:
(if there's no objections I'd like to merge this by the end of the week)

JamesWrigley:
>       * `XGM.intensity()` also has units through [pint-xarray](https://pint-xarray.readthedocs.io/en/latest/index.html). I added it for consistency and because it's nice to have the units displayed when printing the `DataArray`, but I have a hard time coming up with a situation where it's actually useful (personally I only use the pulse-resolved values as an I0) and pint-xarray says it's still an experimental package. So I don't feel very strongly about including it.

One argument in favour of including units is so that it can be used automagically by other things, like DAMNIT.

JamesWrigley:
(if there's no objections I'd like to merge this tomorrow)

tmichela (approved):
LGTM, thanks!

I'm a bit afraid that some useres may be confused about quantity objects, and wonder if we should have an opt-in/out for that 🤔 but I think I'm fine as it is now.
Maybe you can leave this PR open for a couple more days in case Philipp has an opinion on that?

docs/components/index.md
JamesWrigley: Yes, but what I'm saying is that I'd prefer not to sort the components globally. e.g. if there are `Alpha` and `Zeta` top-level components I'd prefer the ordering to be:
- Alpha
- Zeta
- Pulse patterns...
- Scans...

Rather than:
- Alpha
- Pulse patterns...
- Scans...
- Zeta

tmichela: I find this confusing to be honnest. It might be fine as long as we don't have too many components, but then?
Or should we put all component in subsections (like `diagnostic` for XGM?) + an other flat page listing all components alphabetically?

JamesWrigley: Sure, I don't feel very strongly about that. But I will probably leave it for another PR :)

src/extra/components/xgm.py
```diff
@@ -0,0 +1,526 @@
...
+                Passing 0 always selects the main properties of a device,
+                e.g. `data.intensityTD` instead of the SASE-specific ones.
+        """
+        device = _find_xgm(run, device)
+
+        self._control_source = run[device]
+        self._instrument_source = run[f"{device}:output"]
+
+        self._extra_sases = [sase for sase in [1, 3]
+                             if f"pulseEnergy.numberOfSa{sase}BunchesActual" in self._control_source]
```
tmichela, line 167: Was just reading old data, it looks like the keys have changed as some point. Old proposal (e.g. p700000/r1) do not have `pulseEnergy.numberOfBunchesActual` or `pulseEnergy.numberOfSa[13]BunchesActual` but only `pulseEnergy.nummberOfBrunches` (yes, with 2 `m` in nummber)

tmichela: and there's a version with proper spelling `pulseEnergy.numberOfBunches` (e.g. p002429)

JamesWrigley: Nice, thanks for catching that :rofl: :sandwich: I'll probably get around to fixing it next week.

JamesWrigley: This reminds me of #23.

JamesWrigley: Fixed in c378ea5.

src/extra/components/xgm.py
```diff
@@ -0,0 +1,526 @@
...
+                key = "data.intensityTD"
+            else:
+                key = f"data.intensitySa{pg.value}TD"
+
+            max_pulses = self.max_pulses(sase)
+            # We could use .xarray(roi=...) here, but that fails because of h5py
+            # when max_pulses is 0.
+            intensity = self.instrument_source[key].xarray()[:, :max_pulses]
+            # Assign a pulseId dimension and coordinate
+            intensity = intensity.rename(dim_0="pulseId").assign_coords(pulseId=np.arange(max_pulses))
```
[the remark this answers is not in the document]
JamesWrigley: Renamed it to `pulseIndex` in c378ea5. I might leave integration with the other components to a later PR.

src/extra/components/xgm.py
```diff
@@ -0,0 +1,526 @@
...
+    def _check_sase_arg(self, sase):
+        """Helper function to return the correct PropertyGroup."""
+        if len(self._extra_sases) == 0 and sase is not None:
+            raise RuntimeError("This XGM doesn't have data from any other SASEs")
+        if sase is not None and sase not in [0, 1, 3]:
+            raise RuntimeError(f"Invalid SASE '{sase}', it must be either 0, 1, or 3")
+
+        return PropertyGroup(sase) if sase is not None else self._default_pg
+
+    def intensity(self, sase=None) -> "xarray.DataArray":
```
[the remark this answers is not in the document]
JamesWrigley: Changed in c378ea5, though it looks unnecessarily verbose to me :shrug: But I don't feel very strongly about it.

src/extra/components/xgm.py
```diff
@@ -0,0 +1,526 @@
...
+        if pg not in self._npulses_by_train:
+            if pg == PropertyGroup.MAIN:
+                key = "pulseEnergy.numberOfBunchesActual"
+            else:
+                key = f"pulseEnergy.numberOfSa{pg.value}BunchesActual"
+            self._npulses_by_train[pg] = self._control_source[key].xarray()
+
+        return self._npulses_by_train[pg]
+
+    def max_pulses(self, sase=None) -> int:
```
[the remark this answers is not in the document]
JamesWrigley: Renamed to `get_max_npulses()` in c378ea5.

src/extra/components/xgm.py
```diff
@@ -0,0 +1,526 @@
...
+            intensity.data[intensity.data == 1] = 0
+
+            import pint_xarray
+            intensity = intensity.pint.quantify("uJ")
+
+            self._intensity[pg] = intensity
+
+        return self._intensity[pg]
+
+    def npulses(self, sase=None) -> int:
```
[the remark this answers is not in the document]
JamesWrigley: Added in c378ea5.

JamesWrigley:
Allllrighty let's see if I can get this over the finish line :upside_down_face: Changes in c378ea5:
- Fixed review issues
- I used it for realsies in an MID experiment last year, and based on that experience I removed the `pint-xarray` integration in favour of an explicit `units` attribute. Having the array values be pint quantities ended up just being annoying, and an attribute is just as visible to the user. Having units on the wavelength and pulse energy was very useful though so I left those enabled by default (but now there's a `with_units` argument to disable it).
- Added `XGM.plot()` to generate an overview plot.
- Moved the documentation to a new `Beam properties` page. Mayhaps future components for e.g. attenuators or spectrometers could go in there.

JamesWrigley:
(if there's no objections I'd like to merge this sometime next week)

tmichela:
I generally dislike the `get_` prefix, particularly as this is anyway readonly data so there won't ever be `set_`... but this is only personal preference, so LGTM.

JamesWrigley:
@takluyver, any opinion on the naming?

takluyver:
No strong overarching opinion, but I think it could be more coherent with existing methods on the pulse pattern components, e.g.

- `get_npulses_by_train()` vs. `get_pulse_counts()` (the latter also roughly fits with references to 'data counts' in EXtra-data)
- `get_npulses_is_constant()` compared to `is_constant_pattern()` - maybe -> `is_constant_pulse_count()` or `is_pulse_count_constant()`

This is only going to get harder the more API surface we create. Maybe we should write down some patterns we use & examples of them.

takluyver:
I'd make `.get_doocs_server()` into a `.doocs_server` property - it feels more like other things that are already properties than data that you're 'getting', even if the implementation gets it from data.

(No particular leaning on the other `get_` methods, though)

JamesWrigley:
> * `get_npulses_by_train()` vs. `get_pulse_counts()` (the latter also roughly fits with references to 'data counts' in EXtra-data)
> * `get_npulses_is_constant()` compared to `is_constant_pattern()` - maybe -> `is_constant_pulse_count()` or `is_pulse_count_constant()`

Makes sense, renamed in d81bf83.

@philsmt, consensus seems to be vaguely in favour of no `get_` prefix. If we went without it for the XGM, would you be ok with removing it from the pulse components too? I can make the PR to deprecate/rename them.

philsmt:
In principle yes, let's have a look:

* `select_trains` stays
* `get_pulse_ids` -> `pulse_ids`
* `peek_pulse_ids` stays?
* `get_pulse_mask` -> `pulse_mask`
* `is_constant_pattern` stays
* `get_pulse_counts` -> `pulse_counts`
* `get_pulse_index` -> `pulse_index`? `make_pulse_index`? `build_pulse_index`? I'm a bit torn on this method, as it delivers a very special object in the form of `pd.MultiIndex`, not a generic read function. Probably preferring `build_pulse_index`
* `search_pulse_patterns` stays
* `trains` stays

What do you think?

JamesWrigley:
Yep, that all looks good to me :+1: I'd probably go with `build_pulse_index()` too.

JamesWrigley:
Renamed everything and rebased in 1f9db1f.

src/extra/components/xgm.py
```diff
@@ -0,0 +1,658 @@
...
+    def _check_sase_arg(self, sase):
+        """Helper function to return the correct PropertyGroup."""
+        if len(self._extra_sases) == 0 and sase is not None:
+            raise RuntimeError("This XGM doesn't have data from any other SASEs")
+        if sase is not None and sase not in [0, 1, 3]:
+            raise RuntimeError(f"Invalid SASE '{sase}', it must be either 0, 1, or 3")
+
+        return PropertyGroup(sase) if sase is not None else self._default_pg
+
+    def intensity(self, sase=None):
```
philsmt, line 302: Sorry if I'm late with this, but looking over it again I just realized that this is physically wrong. [Intensity](https://en.wikipedia.org/wiki/Intensity) is defined as power per area, while here we're looking at [pulse] energy.

I realize this term is used colloquially sometimes (and even some properties are wrong, e.g. `photonFlux`, which is also defined differently).

Strong feelings against `pulse_energy`?

src/extra/components/xgm.py
```diff
@@ -0,0 +1,658 @@
...
+
+        run_prefix = self._get_run_prefix()
+        device_label = self._get_device_label(sase)
+        fig.suptitle(f"{run_prefix}{device_label}")
+
+        fig.tight_layout()
+
+        return fig.get_axes()[0]
+
+    def plot_intensity(self, sase=None, ax=None, minimal_title=False):
```
philsmt, line 514: Same here as above, the XGM is not measuring intensity physically, but energy.

PR #56, opened 2023-09-11
Scan bugs

JamesWrigley:
Fixes two bugs I found (see the commits for details):
- Backlash detection would sometimes cause exceptions on scans with noisy motors (3a35fe8)
- Sometimes motors will jitter between two specific values, which can cause `ZeroDivisionError`'s (57e596e)

### DIFF
--- a/src/extra/components/scan.py
+++ b/src/extra/components/scan.py
@@ -199,7 +199,7 @@ class Scan:
                          i * step_size)
                  for i in range(n_steps)]
 
-        motor = np.concatenate(steps)
+        motor = np.concatenate(steps, dtype=np.float64)
 
         from xarray import DataArray
         motor = DataArray(motor,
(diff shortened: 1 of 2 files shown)


src/extra/components/scan.py
```diff
@@ -339,10 +348,11 @@ def _get_motor_steps(self, actual_pos, resolution, min_trains, intra_step_filter
         # Detect backlash at the beginning by comparing the step direction of
         # the first and second steps. If there's backlash they're typically not
         # the same, so we remove the first step if so.
-        first_step = steps[1][0] - steps[0][0]
-        second_step = steps[2][0] - steps[1][0]
-        if np.sign(first_step) != np.sign(second_step):
-            del steps[0]
+        if len(steps) > 3:
```
takluyver, line 351: Should this be `>=`? It looks like we only need 3 values to do the check. I guess most real scans will have quite a few more steps in any case.

JamesWrigley: Ah yes, good catch. Fixed in 73fccbd.

takluyver:
LGTM

PR #100, opened 2024-01-02
Fix handling of trains without pulses

philsmt:
The pulses components, in particular after some rework done in https://github.com/European-XFEL/EXtra/pull/40, did not properly handle trains that do not actually have pulses associated with them. This was mainly due to being overly reliant on constructing all kinds of data from the linear series of pulse IDs, which by nature will not contain empty trains in its index.

Most methods simply ignored trains without pulses even if they shouldn't (`.get_pulse_counts()`, `.get_pulse_mask()`, `.search_pulse_patterns()`) while others could simply break (`.select_trains()`). This MR fixes this behaviour across the boards and changes the tests to contain gaps in the pulse data, both in the sense of no data being recorded as well as recorded trains with no pulses.

### DIFF
--- a/src/extra/components/pulses.py
+++ b/src/extra/components/pulses.py
@@ -56,6 +56,19 @@ class PulsePattern:
         self._pulse_ids = None
 
     def _get_train_ids(self):
+        """Low-level access to train IDs.
+
+        This method may be overriden by any implementation of this class
+        for either performance or if the underlying data may contain
+        train IDs that have no pulse IDs associated with it. The default
+        implementation draws train IDs from the series of pulse IDs, and
+        thus cannot contain trains without pulses. This is particularly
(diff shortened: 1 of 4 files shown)


JamesWrigley (approved):
IANAE, but LGTM!

Minor thing: there's a *lot* of magic indices in the tests :sweat_smile: No need to do it now, but in the future it might be good to either add more comments or explicitly compute the indices to show where they're coming from.

philsmt:
Thanks and fair point, I meant to do that. I added a description of the mock pulse patterns now in 4a191d0

philsmt:
(sneaked in another fix to `DldPulses` to use proper train ID coordinates)

I'll take your LGTM by the end of the day.

JamesWrigley (approved):
LGTM!

PR #103, opened 2024-01-05
Add a DLD component

philsmt:
For various delay line detectors in use at SQS, the offline _calibration_ pipeline is used to digitize the analog signals recorded with digitizers and reconstruct them into proper detector hits in time and space as a virtual detector source. As this kind of data is variable-length by nature, it is saved in sufficiently large fixed-length arrays filled with `np.nan` and compressed away.

While this data format is very efficient for storage and interoperable, it is somewhat difficult to load and work on in a pulse or hit-based manner. Based on observations of existing user code, this component transform the train/pulse/hit-shaped data into a linear `pd.Series` or `pd.DataFrame` with multi indices for convenient access.

It integrates into the existing pulse pattern components, though by default will use the internal pulse information saved as part of the processing result.

### DIFF
--- a/docs/components/dld.md
+++ b/docs/components/dld.md
@@ -1,0 +1,1 @@
+::: extra.components.DelayLineDetector
--- a/src/extra/components/dld.py
+++ b/src/extra/components/dld.py
@@ -1,0 +1,300 @@
+
+from copy import copy
+import re
+
+import numpy as np
+import pandas as pd
+
+
+class DelayLineDetector:
+    """Interface for processed delay line detector data.
+
+    Args:
+        data (extra_data.DataCollection): Data to access DLD data from.
+        detector (str, optional): Source name of the detector, only
+            needed if the data includes more than one.
+        pulses (extra.components.pulses.PulsePattern, optional): Pulse
+            component to pull pulse information. If omitted, it is
+            constructed from the internal trigger information.
+    """
+
+    _instrument_re = re.compile(r'^(\w{3}_\w+_DLD\d\/DET\/\w+):output?$')
+
+    def __init__(self, data, detector=None, pulses=None):
+        if detector is None:
+            # Try to find detector automatically.
+            detector = self._find_detector(data)
+        elif detector.endswith(':output'):
+            # Strip output pipeline if passed
+            detector = detector[:-7]
+
+        self._detector_name = detector
+
+        if detector in data.control_sources:
+            # Some very early data may not have the control source yet.
+            self._control_src = data[detector]
+        else:
+            self._control_src = None
+
+        self._instrument_src = data[f'{detector}:output']
+        self._pulses = pulses
+
+        if self._pulses is None:
+            self._pulses = self.pulses()
+
+    def __repr__(self):
+        return "<{} {}>".format(
+            type(self).__name__, self._detector_name)
+
+    @classmethod
+    def _find_detector(cls, data):
+        """Try to find detector source."""
+
+        detectors = set()
+
+        for source in data.instrument_sources:
+            m = cls._instrument_re.match(source)
+            if m is not None:
+                detectors.add(m[1])
+
+        if len(detectors) > 1:
+            raise ValueError('multiple detectors found, please pass one '
+                             'explicitly:\n' + ', '.join(sorted(detectors)))
+        elif detectors:
+            return detectors.pop()
+
+        raise ValueError('no detector found, please pass one explicitly')
+
+    def _align_pulse_index(self, kd, pulse_dim):
+        """Align pulse index to KeyData.
+
+        Args:
+            kd (extra_data.KeyData): KeyData object to align to.
+            pulse_dim ({pulseId, pulseIndex, time}, optional): Label
+                for pulse dimension, pulse ID by default.
+
+        Returns:
+            (pandas.Index or None): Index from internal pulses component
+                aligned to KeyData object in terms of trains.
+        """
+
+        if (data_counts := kd.data_counts()).sum() == 0:
+            return
+
+        pulse_counts = self._pulses.pulse_counts()
+
+        # Merge both counts in a dataframe joined over train ID.
+        entry_counts = pd.merge(
+            pulse_counts.rename('by_pulses'),
+            data_counts.rename('by_data'),
+            how='outer', left_index=True, right_index=True
+        ).fillna(-1).astype(int)
+
+        # Only consider rows with data counts.
+        entry_counts = entry_counts[entry_counts['by_data'] > 0]
+
+        if (entry_counts['by_pulses'] == -1).any():
+            # Missing one or more trains in the pulse information.
+            raise ValueError('missing pulse information for one or more '
+                             'trains with data')
+
+        if len(entry_counts) < min(len(data_counts), len(pulse_counts)):
+            # Missing counts in one or more trains in data, select data
+            # and pulses down.
+            from extra_data import by_id
+            train_sel = by_id[entry_counts.index.to_numpy()]
+
+            pulses = self._pulses.select_trains(train_sel)
+            kd = kd.select_trains(train_sel)
+        else:
+            pulses = self._pulses
+
+        return pulses.build_pulse_index(pulse_dim)
+
+    def _build_reduced_pd(self, data, index, entry_level):
+        """Get variable-length data as series or dataframe.
+
+        Reduce data with a variable number of actual entries among a
+        fixed number of potential entries to build a linear pandas
+        Series (for scalar floating types) or DataFrame (for structured
+        data containing at least one floating type). The floating type
+        is necessary for NaN determination of valid entries.
+
+        The data must be 2-dimensional, with the first axis is assumed
+        to be an entry ("pulse") dimension and the second to only
+        contain actual data up to a variable number for each entry.
+
+        Args:
+            data (ndarray): 2D data array.
+            index (pd.Index or None): Index to apply to reduced data,
+                or None to indicate no valid entries.
+            entry_level (str or None): Additional index level inserted
+                for each entry or omitted if None.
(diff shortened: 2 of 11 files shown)


philsmt:
Reminder to myself: Add pulse height information to edges, if available.

philsmt:
@JamesWrigley Could I ask you as well to have a brief look? It's been dangling for a bit now, and I would like to base more components off it.

src/extra/components/dld.py
```diff
@@ -0,0 +1,304 @@
...
+        return self.get_pulses().get_triggers()
+
+    def get_edges(self, channel_index=True, pulse_dim='pulseId'):
+        """Get raw edges as series or dataframe.
+
+        Args:
+            channel_index (bool, optional): Whether to insert the edge
+                channel as index level and return Series object
+                (default), or as column and return DataFrame object.
+            pulse_dim ({pulseId, pulseIndex, time}, optional): Label
```
JamesWrigley, line 250: I'm hazy on the pulse naming plans, but is the idea that eventually `pulseId` will be something different from `pulseIndex`? Because if they mean the same thing I think we should stick to a single name.

docs/components/dld.md
```diff
@@ -0,0 +1 @@
+::: extra.components.DelayLineDetector
```
JamesWrigley, line 1: This file should be added to the nav section in `mkdocs.yml`: https://github.com/European-XFEL/EXtra/blob/master/mkdocs.yml#L19
And the component index: https://github.com/European-XFEL/EXtra/blob/master/docs/components/index.md

JamesWrigley (changes requested):
I should say that I'm completely unfamiliar with DLD's, so I may have missed something :eyes:

src/extra/components/dld.py
philsmt: Yes, `pulseId` is the index in the bunch pattern table and something we try to establish as a *global* pulse ID across the facility. `pulseIndex` is an enumeration for any beamline's pulses.

docs/components/dld.md
philsmt: Thanks, done in a534bb0

src/extra/components/dld.py
JamesWrigley: Hmm, kinda annoying since there's a lot of code already using `pulseId` but ok :frowning_face: But anyway, do I understand correctly that this is capable of retrieving the edges labelled by either `pulseId` or `pulseIndex` which will be different numbers but refer to the same pulse? If so, what does `time` mean? Is that something coming from the DLD?

philsmt: Unfortunately yes, all kind of equipment has historically just made up its own pulse IDs. The big detectors are particularly big offenders, as they're not even enumerating pulses, but their own frames. Some experiments take frames between pulses, making the mess complete.

When writing the pulse components, I found it hard to use _id_ for something that's arbitrarily enumerated, and thus chose _index_.

These arguments are directly handed off to `PulsePattern.build_pulse_index`, and yes, the aim is to allow different ways how to label pulses. `time` is just that, the actual time of the pulse relative to the beginning of the train. The idea behind this is that some experiments (e.g. at HED) use pulse trains as a means of probing the same at different _times_.

I will extend the documentation a bit more to illuminate this, thanks! 🙂

JamesWrigley (approved):
LGTM!

src/extra/components/dld.py
```diff
@@ -0,0 +1,304 @@
...
+            type(self).__name__, self._detector_name)
+
+    @classmethod
+    def _find_detector(cls, data):
+        """Try to find detector source."""
+
+        detectors = set()
+
+        for source in data.instrument_sources:
+            m = cls._instrument_re.match(source)
```
takluyver, line 56: For sources coming from Karabo, we're trying to use the device class, with the instrument pattern just as a fallback. What do you think about doing something similar here, introducing some kind of 'source type' metadata for the sources we create in calibration?

Obviously this doesn't need to be part of this PR - I'm thinking about the future.

src/extra/components/dld.py
```diff
@@ -0,0 +1,304 @@
...
+
+    @property
+    def instrument_source(self):
+        """Instrument source."""
+        return self._instrument_src
+
+    @property
+    def rec_params(self):
+        """Reconstruction parameters."""
+        return self.control_source.run_values()
```
takluyver, line 206: I just noticed that for REMI, you're writing the `RUN` values without the value & timestamp pairs that we normally see in RUN or CONTROL data. Whereas for the JUNGFRAU ROI reductions, we put the parameters in RUN using your `DataFile` API, which does create value & timestamp datasets.

Do we consider that a bug? If we do and we want to fix it, we might pre-emptively account for the difference here, i.e. discard the `.timestamp` keys and remove the `.value` suffix from the others.

(The mockdata framework will create timestamp & value pairs, so that's also not fully representative of the normal files)

takluyver:
I found the `_get_reduced_pd` method not especially easy to read. It feels like it's trying to cover a lot of possibilities for something that's only called in 3 places.  I suspect that it could be broken up a bit, but I haven't understood it well enough to suggest how. That's not really important if you're happy with it, though.

I think my question about RUN values is worth thinking about before we merge this.

src/extra/components/dld.py
philsmt: Yes, this is definitely a bug. The `.timestamp` part quite often eludes me, as I rarely if ever used it in the field. Thanks for spotting.

src/extra/components/dld.py
philsmt: Yes definitely, I omitted it here as essentially all data is legacy data for now.

In an upcoming PR on another component, I started abstracting all this search-sources-based-on-device-class-or-needle functionality.

takluyver:
No, I don't feel strongly about it. :slightly_smiling_face:

philsmt:
@takluyver I ended up giving it another try to reduce the complexity of `_get_reduced_pd` by splitting it along the messy "it-may-be-a-list-or-not" line. Is this more readable and/or understandable?

src/extra/components/dld.py
```diff
@@ -0,0 +1,389 @@
...
+        contain actual data up to a variable number for each entry.
+
+        The data type must either be a scalar floating type (returning
+        Series objects) or a structured data type (returning DataFrame
+        objects) containing at least a single floating type.
+
+        Args:
+            key (str): Instrument source key.
+            pulse_dim (str): Pulse to use in index, must be supported by
+                underlying pulses object.
```
takluyver, line 134: These two parameters don't match the actual signature.

src/extra/components/dld.py
```diff
@@ -0,0 +1,389 @@
...
+        index_df = index.repeat(
+            finite_mask.reshape(-1, num_rows).sum(axis=1)).to_frame()
+
+        # Insert additional index level enumerating actual entries.
+        if entry_level is not None:
+            index_df[entry_level] = np.flatnonzero(finite_mask) % num_rows
+
+        return pd_cls(raw[finite_mask], pd.MultiIndex.from_frame(index_df))
+
+    def _get_reduced_pd(self, key, pulse_dim, entry_level, split_axis):
```
takluyver, line 175: I think this method is no longer called anywhere?

src/extra/components/dld.py
```diff
@@ -0,0 +1,389 @@
...
+        pulse_counts = self._pulses.pulse_counts()
+
+        # Merge both counts in a dataframe joined over train ID.
+        entry_counts = pd.merge(
+            pulse_counts.rename('by_pulses'),
+            data_counts.rename('by_data'),
+            how='outer', left_index=True, right_index=True
+        ).fillna(-1).astype(int)
+
+        print(entry_counts)
```
takluyver, line 93: Leftover debugging?

src/extra/components/dld.py
```diff
@@ -0,0 +1,389 @@
...
+            (pd.Series or pd.DataFrame): Raw edge positions as Series
+                (indexed by channel) or DataFrame (with channel as
+                column) object.
+        """
+
+        kd = self._instrument_src['raw.edges']
+        index = self._align_pulse_index(kd, pulse_dim)
+        data = kd.ndarray()
+        raw_edges = [self._build_reduced_pd(data[:, i, :], index, None)
+                     for i in range(7)]
```
takluyver, line 342: Without knowing anything about it, the hardcoded 7 feels a bit odd.  Are we confident it can never change? Does `data.shape[1]` work?

takluyver (commented):
Thanks, that does look easier to read now.

src/extra/components/dld.py
philsmt: Certainly, for comparison with the new version.

src/extra/components/dld.py
philsmt: 🙈

src/extra/components/dld.py
philsmt: Oops, forgot to update the docstring here.

src/extra/components/dld.py
philsmt: Essentially for the moment yes, as it's hardcoded in the REMI/DLD notebook. But your suggestion makes sense, thanks!

src/extra/components/dld.py
philsmt: Would you then drop `.value` or *simulate* it? I gravitate towards the former (and I wanted to add a keyword argument to `.run_values()` for a while now to facilitate that btw).

philsmt: I made up my mind, since it doesn't openly imply it comes from `RUN` and dating it is kind of pointless (you'd want the modification date of the underlying YAML file now, whatever it is later).

philsmt:
Thanks for review!

PR #106, opened 2024-01-22
Deprecate the `get_` prefix from some method names

JamesWrigley:
See: https://github.com/European-XFEL/EXtra/pull/53#issuecomment-1903955087

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -28,6 +28,8 @@ Added:
 
 Changed:
 
+- The `get_` prefix was deprecated for some method names in the [pulse pattern
+  components](components/pulse-patterns.md) (!106).
 - All methods in [XrayPulses][extra.components.XrayPulses] and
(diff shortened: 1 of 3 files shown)


philsmt:
LGTM, thanks :pray: 

Are we removing those for the next run (i.e. 2024/2)?

JamesWrigley:
Yeah, I think next run is reasonable since I assume there's not too many uses of this in the wild yet :+1:

tmichela:
you mean current run (202401), right?

philsmt:
Now, I mean removing it at the beginning of 2024/2. It seems prudent to have the deprecation warning to at least appear for a run.

JamesWrigley:
(or even a year if the deprecated thing is widely used, since many users only come around once a year)

tmichela:
👍

PR #108, opened 2024-01-23
Rename 'intensity' to 'energy'

JamesWrigley:
This is more physically accurate since the XGM actually gives back an energy rather than intensity in its `data.intensityTD` property.

I also renamed the plot methods to `XGM.plot_energy_per_pulse()` and `XGM.plot_energy_per_train()`. And fixed an embarrassing docstring bug claiming that `XGM.pulse_energy()` returns things in keV :flushed:


philsmt:
Thanks, LGTM

JamesWrigley:
I think we're ready to make a release :) I leave that to you since I think you own the pypi package.

PR #111, opened 2024-01-25
Add preview CalCat API

takluyver:
Example usage:

```python
from extra.calibration import CalibrationData, LPDConditions

lpd_cd = CalibrationData.from_condition(
    LPDConditions(memory_cells=200, sensor_bias_voltage=250),
    "FXE_DET_LPD1M-1",
    event_at="2022-05-22T02:00:00",
)
lpd_cd['Offset'].ndarray()
```

This is still quite rough and will probably need additions. It has tests which are running & passing internally (in pycalibration), but at present they all rely on the internal network to access exflcalproxy, so I haven't copied them here yet.


philsmt:
Since there's a bit of docstrings available, would it make sense to add a doc entry? Or would you draw the line there due to its preview status?

philsmt:
LGTM otherwise

takluyver:
I might hold off on adding it to the docs until we've been able to play with it a bit in pratice. :shrug:

PR #119, opened 2024-02-07
Add some tests for extra.calibration

takluyver:
If I've set everything up right, this much should work on Github actions. I've left out the tests that try to actually load data.


takluyver:
Huh, it actually did work.

JamesWrigley:
What do you think about disabling the tests for this by default, but explicitly enabling them in CI? The reason is because people (me :stuck_out_tongue: ) who run the tests locally will need to explicitly disable them every time if they haven't got the right token set up.

takluyver:
I think that's an excellent idea... :grin: 

https://github.com/European-XFEL/EXtra/blob/9587d702d80ecde3cdf16d69ec33fdb921a0a2a6/tests/test_calibration.py#L17-L18

JamesWrigley:
Looool I can't read :upside_down_face:

takluyver:
I've reworked this, so on Github it tests with saved responses from CalCat, using the [pytest-recording](https://github.com/kiwicom/pytest-recording) plugin. This should catch any issues coming from changes in the code, and it works without any secrets. This is also what will happen when we run the tests manually.

Once this is merged, I'll also set up some nightly tests in Gitlab which will do the same thing with new responses from the server, to catch problems that might be caused by server changes.

PR #124, opened 2024-02-20
Add support for binning data to the Scan component

JamesWrigley:
This is usually the entire point of a scan so everyone will end up doing this anyway. I also took the liberty of cleaning up the changelog a bit in 9c39677.

One thing I wasn't sure about: some people prefer the standard error instead of the standard deviation. Should we use that by default? Or make it an option?


src/extra/components/scan.py
```diff
@@ -163,6 +162,83 @@ def plot(self, ax=None):
 
         return ax
 
+    def bin(self, data):
```
philsmt, line 165: I would prefer a more expressive name here, as it initially confused me what is actually binned here to what. On top of that, I'm not a fan of collisions with builtins.

As I understand, this method can take arbitrary train-resolved data and averages it for each bin this scan represents. Binning in itself doesn't mean averaging to me, it just means combining results based on some metric. What do you think of explicitly calling it `.step_averages()`?

JamesWrigley: Sure, renamed it to `step_average()` in dd35a73 (`average_by_steps()` is a bit long methinks).

src/extra/components/scan.py
```diff
@@ -163,6 +162,83 @@ def plot(self, ax=None):
...
+
+        return xr.DataArray(signal, dims=("position",),
+                            coords={"position": ("position", self.positions),
+                                    "stdev": ("position", stdev)},
+                            name=data.name,
+                            attrs={
+                                "motor": self.name
+                            })
+
+    def plot_bin(self, data, title=None, xlabel=None, ylabel=None,
```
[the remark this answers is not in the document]
JamesWrigley: Renamed to `plot_step_average()` in dd35a73.

philsmt (changes requested):
No comments on the actual implementation, but I would prefer some less ambiguous naming. Really cool functionality!

I'm already wondering what other reduction may be useful next to average. Maybe that could also be a naming direction...

philsmt (approved):

JamesWrigley:
How about `bin_steps()` or `reduce_steps()`? That would leave the door open for different reductions in the future.

philsmt:
> `bin_steps()` or `reduce_steps()`

Somehow these suggest something is done **to** steps, rather than the steps being used to sort/bin/reduce/average something **else**. But this may simply be due to English being a second language.

Maybe Thomas has a good idea when he's back. But I'd also be fine to use the existing names for average here, and later implement it as a specific case of a more general function.

JamesWrigley:
Yeah, I'm using it more loosely in a 'bin the data that was collected during these steps' way. But I don't feel very strongly about it, how about `bin_by_steps()`?

philsmt:
> how about `bin_by_steps()`?

I didn't dare to suggest that, but it strongly communicates the meaning to me :see_no_evil:

JamesWrigley:
Alrighty, renamed to `bin_by_steps()` in 8a6a5dc.

JamesWrigley:
Ok one more naming bikeshed :upside_down_face: Is it confusing that the label on the plots is `Error: ...`? Like do you think people will confuse that with an actual error from the code? If so then we could call it 'uncertainty' instead.

philsmt:
Ah yes, error is indeed a loaded term. I spoke to Danilo, and he strongly preferred the term "Statistical uncertainty" instead. In the docs, we should explicitly mention it's the 68% confidence interval (also called 1-sigma region).

JamesWrigley:
Roger that, renamed everything to 'uncertainty' in f0fc09f.

JamesWrigley:
(I will sneakily take your earlier LGTM :smirk: )

PR #125, opened 2024-02-20
Add PumpProbePulses to the component index

JamesWrigley:
Also reordered the components in the documentation such that they vaguely go down by order of 'specificity'.

I noticed it was missing while reviewing #103.


philsmt (approved):
Oops, thanks. LGTM.

PR #126, opened 2024-02-23
Fix PumpProbePulses.pulse_mask() for empty or partially empty trains

philsmt:
@bj-s reported an issue with `PumpProbePulses.pulse_mask()` in a run where all trains are pumped, but every 50th train has no FEL pulses.

The special implementation of `pulse_mask()` implicitly assumed every train to have at least one FEL and PPL pulse in its implementation. Additionally while fixing this, I noticed it was not included in the fixes in https://github.com/European-XFEL/EXtra/pull/100 regarding handling of trains without pulses.

Extended the tests to cover what previously failed.


JamesWrigley (approved):
LGTM!

philsmt:
Thanks for review!

PR #127, opened 2024-02-23
Store & access all CCV metadata from CalCat

takluyver:
I stripped almost all the metadata attributes from `SingleConstant`, because I think the API should be usable even without talking to CalCat. But naturally there are times when we want that metadata, e.g. to show the timestamp (`begin_validity_at`) in correction reports.

This is designed to handle 3 scenarios:

1. We found the constants from CalCat - store all the CCV metadata from the response, make it available 'free'
2. We got constants another way which includes CCV IDs, e.g. metadata from a correction - you can use constants without talking to CalCat, but if you want the metadata, it will need 1 query per SingleConstant object.
3. We got constants without CCV IDs, e.g. testing constants that have not yet been injected to CalCat - no metadata available, but you can still organise them & load data through this API.

### DIFF
--- a/src/extra/calibration.py
+++ b/src/extra/calibration.py
@@ -256,6 +256,7 @@ class SingleConstant:
     dataset: str
     ccv_id: Optional[int]
     pdu_name: Optional[str]
+    _calcat_metadata: Optional[dict] = None
 
     @classmethod
     def from_response(cls, ccv: dict) -> "SingleConstant":
@@ -264,6 +265,7 @@ class SingleConstant:
             dataset=ccv["data_set_name"],
             ccv_id=ccv["id"],
             pdu_name=ccv["physical_detector_unit"]["physical_name"],
+            _calcat_metadata=ccv,
         )
 
     def dataset_obj(self, caldb_root=None) -> h5py.Dataset:
@@ -277,6 +279,17 @@ class SingleConstant:
 
     def ndarray(self, caldb_root=None):
         return self.dataset_obj(caldb_root)[:]
+
+    def calcat_metadata(self, client=None):
+        if self._calcat_metadata is None:
+            if self.ccv_id is None:
+                raise Exception("Unable to retrieve metadata without ccv_id")
+
+            client = client or get_client()
+            self._calcat_metadata = client.get(
+                f"calibration_constant_versions/{self.ccv_id}"
+            )
+        return self._calcat_metadata
 
 
 def prepare_selection(
--- a/src/extra/calibration.py
+++ b/src/extra/calibration.py
@@ -1,7 +1,7 @@
 import json
 import re
 from collections.abc import Mapping
-from dataclasses import dataclass, replace
+from dataclasses import dataclass, field, replace
 from datetime import date, datetime, time, timezone
 from functools import lru_cache
 from pathlib import Path
@@ -256,7 +256,8 @@ class SingleConstant:
     dataset: str
     ccv_id: Optional[int]
     pdu_name: Optional[str]
-    _calcat_metadata: Optional[dict] = None
+    _metadata: dict = field(default_factory=dict)
+    _have_calcat_metadata: bool = False
 
     @classmethod
     def from_response(cls, ccv: dict) -> "SingleConstant":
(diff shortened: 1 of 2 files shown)


src/extra/calibration.py
```diff
@@ -278,6 +280,17 @@ def dataset_obj(self, caldb_root=None) -> h5py.Dataset:
     def ndarray(self, caldb_root=None):
         return self.dataset_obj(caldb_root)[:]
 
+    def calcat_metadata(self, client=None):
```
philsmt, line 283: Name bikeshedding inc.

Mind just using `metadata()`? I could imagine this also being part of some better constant format, or having more metadata there over time.

src/extra/calibration.py
```diff
@@ -278,6 +280,17 @@ def dataset_obj(self, caldb_root=None) -> h5py.Dataset:
...
+        if self._calcat_metadata is None:
+            if self.ccv_id is None:
+                raise Exception("Unable to retrieve metadata without ccv_id")
```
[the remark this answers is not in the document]
takluyver: Yeah, I was using `Exception` as a placeholder while we discussed what it should be. I also wondered a bit about creating a specific exception class for this, but that might be overkill - I can't think of many scenarios where you'd actually need to catch it.

src/extra/calibration.py
takluyver: My thinking with putting 'calcat' in there was to give a hint that a) it might not be available if the constants you're looking at aren't in calcat, and b) it might require a request to calcat to get the information.

But it's a good point that some or all of this metadata may be available outside calcat. Even if we load `calibration_metadata.yml` from a correction, we have the begin validity timestamp, and it would be nice to keep that available even without the rest of the metadata. :thinking:

JamesWrigley (approved):
LGTM!

On a side note, is this ready to be documented?

src/extra/calibration.py
```diff
@@ -278,6 +282,37 @@ def dataset_obj(self, caldb_root=None) -> h5py.Dataset:
...
+
+    def metadata_dict(self, client=None):
+        """Get a dict of available metadata
+
+        If this constant didn't come from CalCat but we have a CalCat CCV ID,
+        this will fetch metadata from CalCat.
+        """
+        if (not self._have_calcat_metadata) and (self.ccv_id is not None):
+            self._load_calcat_metadata(client)
+        return self._metadata.copy()
```
philsmt, line 314: So this returns the full metadata if there's a CCV ID (whatever it was initialized with + what's in CalCat), but only what it was initialized with in other cases?

tests/test_calibration.py
```diff
@@ -57,6 +57,10 @@ def test_AGIPD_CalibrationData_metadata():
     assert isinstance(agipd_cd["Offset", "AGIPD00"], SingleConstant)
     assert agipd_cd["Offset", "Q1M2"] == agipd_cd["Offset", "AGIPD01"]
 
+    bva = agipd_cd["Offset", "AGIPD00"].metadata("begin_validity_at")
```
philsmt, line 60: It seems to me these tests do not check the "no-CCV-ID" case for both `.metadata()` and `.metadata_dict()`. Is this state _currently_ possible, or would it only happen for potential future cases (you mentioned `calibration_metadata.yml`)?

src/extra/calibration.py
takluyver: Yup, that's right.

tests/test_calibration.py
takluyver: It's basically a potential future scenario - e.g. if you've just created a new batch of constants in a folder, and you want to make a `CalibrationData` object to look at them before injecting them to CalCat. So my aim is that the design won't need breaking changes if/when we do that, rather than making that work now.

philsmt:
Thanks, LGTM

takluyver:
Thanks both!

Regarding the docs: IDK. We've designed a fairly big new piece of API surface with only a very little bit of usage of one prototype, so I'd quite like to get a bit more use (internal or external) before we document it and get more reluctant to change things. But it has been hanging around for months already, and using it internally for offline calibration is stuck behind upgrading offline calibration to a newer Python. Maybe we should copy the version of the module in EXtra back into the internal calibration tools to try it out sooner.

PR #129, opened 2024-02-26
Make CI check that the changelog was modified

JamesWrigley:
It can be overridden with the `skip-changelog` label.

### DIFF
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -1,0 +1,25 @@
+# Check that the changelog has been updated.
+# See: https://github.com/marketplace/actions/changelog-enforcer
+
+name: "PR workflow"
+on:
+  pull_request:
+    # The specific activity types are listed here to include "labeled" and "unlabeled"
+    # (which are not included by default for the "pull_request" trigger).
+    # This is needed to allow skipping enforcement of the changelog in PRs with specific labels,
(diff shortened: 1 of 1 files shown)


JamesWrigley:
Tested this by opening the PR without the `skip-changelog` label (it failed), and then adding the label (it succeeded).

takluyver:
My guess is that this would result in a lot of merge conflicts, as PRs overlap in time and would all have to update the changelog in the same place. They'd be easy conflicts to merge, so that's not necessarily a big problem, but it's not ideal.

Quite a few projects work with a directory of news fragments - you add a fragment file with each PR, and then a tool pulls them together at release time. [Towncrier](https://towncrier.readthedocs.io/en/stable/) is the most popular tool I know of specifically for that, but in h5py we use [rever](https://regro.github.io/rever-docs/), which includes this functionality, and I think custom scripts are not uncommon. For EXtra-data, I've been trialling this approach where you add a 'prnote:' comment on the pull request, which feels 'lighter' than adding a file in the repo.

If we're happy just dealing with the merge conflicts, though, the actual implementation of the check looks fine to me.

JamesWrigley:
New motivation for quick reviews: avoiding merge conflicts from other PRs :trollface: 
Maybe we could try it for a bit and see how it works? Hopefully the repo isn't high-traffic enough to create many merge conflicts.

PR #131, opened 2024-02-29
Implement fit_gaussian() to quickly fit a Gaussian

JamesWrigley:
I've been copying this around for a few years, it's quite handy to quickly fit something.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -15,6 +15,9 @@
     ```bash title="Installation command"
     pip install git+https://github.com/European-XFEL/EXtra.git
     ```
+Added:
+
+- A helper function named [fit_gaussian()][extra.utils.fit_gaussian] (!131).
 
 ## [2024.1.1]
 
--- a/src/extra/utils/functions.py
+++ b/src/extra/utils/functions.py
@@ -23,6 +23,59 @@ def gaussian(x, y0, A, μ, σ):
     return y0 + (A / (σ * np.sqrt(2*np.pi))) * np.exp(-(x - μ)**2 / (2 * σ**2))
 
 
+def fit_gaussian(ydata, xdata=None, p0=None, **kwargs):
+    """Fit a Gaussian to some data.
+
+    This uses [curve_fit()][scipy.optimize.curve_fit] to fit a Gaussian (from
+    [gaussian()][extra.utils.gaussian]) to `ydata`. If `p0` is not passed the
+    function will set them to reasonable defaults. It will return `None` if
+    fitting fails, or if there are no non-NaN values in `ydata`.
+
+    !!! note
+        By default this will only return the `popt` array from
+        [curve_fit()][scipy.optimize.curve_fit], if you want `pcov` or any other
+        output you must pass `full_output=True`.
+
+    Args:
+        ydata (array_like): The data to fit. NaN's will automatically be masked
+            before fitting.
+        xdata (array_like): Optional x-values corresponding to `ydata`.
+        p0 (list): A list of `[y0, A, μ, σ]` to match the arguments to
+            [gaussian()][extra.utils.gaussian].
+        **kwargs (): All other keyword arguments will be passed to
+            [curve_fit()][scipy.optimize.curve_fit].
+    """
+    if xdata is None:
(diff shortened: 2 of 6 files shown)


src/extra/utils/functions.py
```diff
@@ -23,6 +23,53 @@ def gaussian(x, y0, A, μ, σ):
...
+        A = max(1, ydata[μ_idx])
+        σ = abs(np.max(xdata) - np.min(xdata)) / 4
+        y0 = np.min(ydata)
+        p0 = [y0, A, μ, σ]
+
+    try:
+        result = curve_fit(gaussian, xdata, ydata, p0=p0)
+        return result if return_all else result[0]
+    except RuntimeError:
+        return None
```
takluyver, line 70: It seems like it would be easy to assume this function succeeds, do something with the results like `y0, A, μ, σ = fit_gaussian(...)` and get a confusing TypeError if the fit fails.

Unless there's a good reason why returning None is helpful, I think it's probably better to just let the RuntimeError propagate.

src/extra/utils/functions.py
```diff
@@ -23,6 +23,53 @@ def gaussian(x, y0, A, μ, σ):
     return y0 + (A / (σ * np.sqrt(2*np.pi))) * np.exp(-(x - μ)**2 / (2 * σ**2))
 
 
+def fit_gaussian(ydata, return_all=False, xdata=None, p0=None, **kwargs):
```
takluyver, line 26: Signature nitpicking :nerd_face: :

`return_all` seems confusingly similar to the `full_output` parameter in `curve_fit()`, but you'd have to specify both `return_all=True, full_output=True` to get all the possible output. Can we make do with just one parameter, preferably following scipy's name?

It also seeems odd to me that this comes between ydata & xdata. I think passing booleans as positional arguments usually makes code less clear, unless there's a really obvious meaning:

```python
fit_info = fit_gaussian(arr, True)   # What does True mean?

fit_info = fit_gaussian(arr, full_output=True)  # Clearer
```

So I might make the option keyword-only.

src/extra/utils/functions.py
```diff
@@ -23,6 +23,53 @@ def gaussian(x, y0, A, μ, σ):
...
+    """
+    if xdata is None:
+        xdata = np.arange(len(ydata))
+
+    from scipy.optimize import curve_fit
+
+    # Filter nans
+    nan_mask = ~np.isnan(ydata)
+    xdata = xdata[nan_mask]
+    ydata = ydata[nan_mask]
```
takluyver, line 52: It looks like recent versions of scipy have a `nan_policy` parameter, where `'omit'` is one option. However our 202401 environment still has scipy 1.8, so that's a question for another day.

src/extra/utils/functions.py
JamesWrigley: I actually prefer returning `None` because my usual use-case for this is fitting entire runs of data per-train where it's normal for the fitting to occasionally fail, and I'd rather this return `None` than having to remember the right exception to catch every time.

src/extra/utils/functions.py
JamesWrigley: Yes, both of those make sense :+1:

src/extra/utils/functions.py
takluyver: OK, fair enough.

src/extra/utils/functions.py
JamesWrigley: Fixed in bc05cbf.

docs/changelog.md
```diff
@@ -20,6 +20,7 @@
 
 Added:
 
+- A helper function named [fit_gaussian()][extra.utils.fit_gaussian] (!131).
```
takluyver, line 23: Can you move this up into the 'unreleased' section?

(It was probably in the right place when you made the PR, but I inserted a subheading to do a release, so now it's no longer correct.

takluyver:
Other than that, LGTM.

docs/changelog.md
JamesWrigley: Oops, yes, fixed in 2af0c8d.

PR #134, opened 2024-03-05
Add conditions object for finding JUNGFRAU constants

takluyver:
I have copied the details across from [the internal calcat_interface code](https://git.xfel.eu/calibration/pycalibration/-/blob/881bb802824079fa54c9dcbbf65e173ffb733155/src/cal_tools/calcat_interface.py#L1175-1238), but it would be good to double check that.

cc @kakhahmed

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -24,6 +24,7 @@ Added:
 - `pulse_periods()`, `pulse_repetition_rates()` and `train_durations()` methods
   to obtain statistics about the pulses in all [pulse
   pattern](components/pulse-patterns.md) components (!114).
+- `extra.calibration` module to find and load calibration constants.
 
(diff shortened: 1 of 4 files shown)


kakhahmed:
LGTM.
It's good to know about these cassettes. I guess there is no strategy for updating these cassettes as updating the response is something we control in the CAL team. Or do you think otherwise?

takluyver:
Thanks!

We've got [tests running daily](https://git.xfel.eu/dataAnalysis/extra-tests/-/pipelines) against CalCat, ignoring the cassettes. So if there's a change in the responses which makes the tests fail, we'll pick it up there, write a fix and re-record the relevant cassettes. There's no plan so far for picking up changes in the responses which don't cause test failures.

PR #135, opened 2024-03-05
Add docs for extra.calibration module

takluyver:
Both a page in the docs, and docstrings for a lot of pieces that didn't yet have them. Let's see how this looks once built.


JamesWrigley:
BTW it's usually faster to build the docs locally with `mkdocs serve` instead of waiting for CI: https://github.com/European-XFEL/EXtra?tab=readme-ov-file#development

takluyver:
Thanks, I did get that set up after the initial commit. :slightly_smiling_face:

takluyver:
Merging on the grounds that any docs are good - we can always refine them later.

PR #138, opened 2024-03-08
Add MachinePulses implementation for arbitrary pulse pattern information

philsmt:
The final member in the originally planned `PulsePattern` family: `MachinePulses`

It is essentially a more general version of `XrayPulses` and `OpticalLaserPulses`, allowing to access arbittary bit masks in the pulse pattern table. By default, it is configured to look at all electron bunches (_pulses_) as means to investigate machine operation. This is also the only mode supported when used with pulse pattern decoder data.

When used with timeserver data, any combination of bits can be accessed with both all (`mode='and'`) or any (`mode='or'`, default) bits set. 

There a couple of small changes to parent classes to accomodate the previously unplanned support of ppdecoder at all.

### DIFF
--- a/src/extra/components/pulses.py
+++ b/src/extra/components/pulses.py
@@ -9,7 +9,9 @@ import re
 
 import numpy as np
 
-from euxfel_bunch_pattern import PPL_BITS, is_sase, is_laser
+from euxfel_bunch_pattern import is_sase, is_laser, \
+    PPL_BITS, DESTINATION_TLD, DESTINATION_T4D, DESTINATION_T5D, \
+    PHOTON_LINE_DEFLECTION
 from extra_data import SourceData, KeyData, by_id
 
 from .utils import identify_sase
@@ -28,7 +30,8 @@ except ImportError:
         return zip(a, b)
 
 
-__all__ = ['XrayPulses', 'OpticalLaserPulses', 'PumpProbePulses', 'DldPulses']
+__all__ = ['XrayPulses', 'OpticalLaserPulses', 'MachinePulses',
+           'PumpProbePulses', 'DldPulses']
 
 
 def _drop_first_level(pd_val):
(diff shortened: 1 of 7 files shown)


src/extra/components/pulses.py
```diff
@@ -940,15 +959,111 @@ def _identify_ppl_seed(cls, data):
...
     @property
     def ppl_seed(self) -> Optional[PPL_BITS]:
         """Used laser seed."""
         return self._ppl_seed
 
 
+class MachinePulses(TimeserverPulses):
+    """An interface to machine pulses and other bunch pattern data.
+
+    The bunch pattern table contains many more information than the
```
JamesWrigley, line 974: Nitpicking:
```suggestion
    The bunch pattern table contains much more information than just the
```

src/extra/components/pulses.py
```diff
@@ -940,15 +959,111 @@ def _identify_ppl_seed(cls, data):
...
+        data (extra.data.DataCollection): Data to access bunch pattern
+            data from.
+        source (str, optional): Source name of a timeserver or pulse
+            pattern decoder, only needed if the data includes more than
+            one such device or none could not be detected automatically.
+        mask (int, optional): Custom mask into the bunch pattern table,
+            by default all electron dumps are used. When initialized
+            with ppdecoder data, only the main and SASE dumps are
+            supported.
+        mode (`or` or `and`, optional): Whether to consider entries
```
JamesWrigley, line 999: What do you think about `require_entry_bits="any"` and `require_entry_bits="all"`? To me that's a bit more readable than `mode="or`/`mode="and`.

tests/mockdata/timeserver.py
```diff
@@ -1,21 +1,33 @@
...
+from euxfel_bunch_pattern import DESTINATION_TLD, DESTINATION_T4D, \
+    DESTINATION_T5D, PHOTON_LINE_DEFLECTION, PPL_BITS, \
+    is_destination, is_sase, is_laser
 from extra_data.tests.mockdata.base import DeviceBase
 
 
-def _fill_bunch_pattern_table(table, num_rows, offset=10):
+def _fill_bunch_pattern_table(table, N, offset=10):
+    # The path of electron bunches can be traced via the dump it ends up
+    # in. Generally all bunches end up in the main buch, except those
```
JamesWrigley, line 12: Nitpicking:
```suggestion
    # in. Generally all bunches end up in the main bunch, except those
```

tests/test_components_pulses.py
```diff
@@ -449,6 +450,53 @@ def test_optical_laser_specials(mock_spb_aux_run):
...
+                               mask=DESTINATION_TLD | PPL_BITS.LP_SPB)
+    np.testing.assert_equal(odd_pulses.pulse_ids().iloc[:50], np.r_[:300:6])
+
+    # Main dump and SA2.
+    odd_pulses = MachinePulses(timeserver_run, mode='and',
+                               mask=DESTINATION_TLD | DESTINATION_T5D)
+    assert odd_pulses.pulse_ids().empty
+
+    # Other types than string.
+    assert (MachinePulses(timeserver_run, mask=str(PHOTON_LINE_DEFLECTION)) \
```
JamesWrigley, line 484: Why are strings allowed as a mask? :thinking: (it's not documented BTW)

JamesWrigley (changes requested):

tests/test_components_pulses.py
philsmt: Oops, the comment meant to read `# Other types than int.`

The point was to test casting the mask to `int` in `MachinePulses.__init__`, which on the implementation side was simply to ensure no weird `TypeError`s. But it's not really intended behaviour to pass `str` as such, only the exception is relevant. Shall I remove that test?

JamesWrigley: I think it's better to explicitly fail if anything other than an int is passed, if that's really the only intended argument type. If we allow strings then people might get confused since these are bit masks and try passing binary strings like `"001001"` without realizing those will be treated as decimal numbers.

philsmt: I remember now why I added the cast, I was afraid of `numpy`'s integer types, which can sometimes do weird things:

```python
In [5]: np.uint64(5) & (1 << 2)
TypeError                                 Traceback (most recent call last)
Cell In[5], line 1
----> 1 np.uint64(5) & (1 << 2)

TypeError: ufunc 'bitwise_and' not supported for the input types, and the inputs could not be safely coerced to any supported types according to the casting rule ''safe''
``` 

I agree your point with strings of non-decimal integers... it would be weird to try, but indeed not produce the expected result. Shall I check against `isinstance(mask, np.integer)` before casting to `int` and raise a `TypeError` in all other cases?

src/extra/components/pulses.py
philsmt: That also sounds quite clunky to me. How about we switch to a boolean flag like `require_all_bits=False`?

philsmt: Or, if it's more intuitive, `require_any_bit=True`

src/extra/components/pulses.py
philsmt: Oops, accidently replied to this comment above.

tests/test_components_pulses.py
JamesWrigley: Yep, that sounds good :+1:

src/extra/components/pulses.py
JamesWrigley: Sure, both of those are quite readable.

src/extra/components/pulses.py
```diff
@@ -940,15 +959,115 @@ def _identify_ppl_seed(cls, data):
...
+        source (str, optional): Source name of a timeserver or pulse
+            pattern decoder, only needed if the data includes more than
+            one such device or none could not be detected automatically.
+        mask (int, optional): Custom mask into the bunch pattern table,
+            by default all electron dumps are used. When initialized
+            with ppdecoder data, only the main and SASE dumps are
+            supported.
+        require_all_bits (bool, optional): Whether to consider entries
+            having any or all of the bits specified by mask, False by
+            default. Only supported when initialized with timeserver.
```
JamesWrigley, line 1001: Nitpicking:
```suggestion
            default. Only supported when initialized with a timeserver.
```

JamesWrigley (changes requested):

src/extra/components/pulses.py
philsmt: Ah, that was supposed to be `with timeserver data.`

JamesWrigley (approved):
LGTM!

philsmt:
Thanks for review!

philsmt:
I was certain I pressed merge earlier this week 🤔

PR #140, opened 2024-03-14
Add job to publish packages on PyPI when a tag passes tests

takluyver:
@daviddoji was using an older version, because the 2024.1 tag was not published to PyPI. This automates that step, so packages will be uploaded automatically for a tag, so long as it passes tests.

@philsmt @tmichela This also needs some config on PyPI to allow it to upload ('trusted publisher'). I don't seem to have enough access to set that up.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -3,6 +3,7 @@ name: Tests
 on:
   push:
     branches: [ master ]
+    tags: [ '*' ]
   pull_request:
 
 jobs:
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -36,3 +36,27 @@ jobs:
         python3 -m pytest .
       env:
         CALCAT_OAUTH_CREDS: ${{ secrets.CALCAT_OAUTH_CREDS }}
+
+  publish:
+    runs-on: ubuntu-latest
+    if: ${{ startsWith(github.ref, 'refs/tags/') }}
+    needs: tests
+    permissions:
+      id-token: write  # OIDC for uploading to PyPI
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Set up Python 3.10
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yml
```diff
@@ -36,3 +36,27 @@ jobs:
         python3 -m pytest .
       env:
         CALCAT_OAUTH_CREDS: ${{ secrets.CALCAT_OAUTH_CREDS }}
+
+  publish:
```
tmichela, line 40: shouldn't this block be indented under `jobs`?

tmichela (commented):
IANAE, so comments mostly based on my ignorance of github workflows.

And, should this be added at the top to run on new tags?

```yaml
on:
  push:
    tags:
      - '\d{4}\.\d+'
```

tmichela:
> @philsmt @tmichela This also needs some config on PyPI to allow it to upload ('trusted publisher'). I don't seem to have enough access to set that up.

Same for me looks like only Philipp is owner

.github/workflows/tests.yml
takluyver: It is, but just with two spaces. :slightly_smiling_face:

takluyver:
Thanks, I missed the on-push-tag thing.

I've copied what we already have for EXtra-data, which has been fine so far, but I don't mind making it more specific if we want. The on-push-tag filter [uses glob patterns](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpushbranchestagsbranches-ignoretags-ignore) rather than regexes, though. And it [doesn't look like](https://docs.github.com/en/actions/learn-github-actions/expressions#operators) there's a `=~` operator, unless it's undocumented.

tmichela:
Fine then, LGTM

takluyver:
Philipp upgraded my access to the project on PyPI, so I've set it up there. With a bit of luck this should all work next time we make a tag.

PR #144, opened 2024-03-15
Update Github actions to fix warnings about Node.js 16

takluyver:
[The warnings](https://github.com/European-XFEL/EXtra/actions/runs/8297056392) say:

> Node.js 16 actions are deprecated. Please update the following actions to use Node.js 20: actions/checkout@v3, actions/setup-python@v4, actions/cache@v3. For more information see: https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -16,14 +16,14 @@ jobs:
 
     steps:
     - name: Checkout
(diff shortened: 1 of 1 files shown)


takluyver:
Checked that the warnings are gone; this doesn't need review.

PR #146, opened 2024-03-22
Calibration: Add source names for detector modules (using source_name_pattern)

takluyver:
CalCat now allows specifying a source_name_pattern for detectors, such as `'HED_DET_AGIPD65K1/DET/{modno+8}CH0:xtdf'`. With this, we can expose source names as yet another way to describe detector modules.


philsmt:
This has been merged as part of `DetectorData`. Would you still like to implement the `.source_names()` method as part of this MR, or a new one?

PR #147, opened 2024-03-25
Mask all non-finite values in fit_gaussian()

JamesWrigley:
Because `curve_fit()` also doesn't like infs.


JamesWrigley:
I'm roguishly gonna go ahead and merge this because it's annoying to work around :see_no_evil:

takluyver:
Retrospective LGTM :wink:

PR #148, opened 2024-03-25
Detector motors component

egorsobolev:
This adds component for motors that move detector quadrants. It supposed to used in `extra_geom` to adjust geometry, see European-XFEL/EXtra-geom#269

cc @takluyver, @JamesWrigley

### DIFF
--- a/src/extra/components/detector_motors.py
+++ b/src/extra/components/detector_motors.py
@@ -52,7 +52,7 @@ class DetectorMotors:
 
         self.shape = tuple(len(v) for v in coordinates.values())
         self.num_sources = np.prod(self.shape)
-        self.dc = dc.select(motors.values())
+        self.dc = dc.select(motors.values(), require_all=True)
 
         self.num_trains = len(self.dc.train_ids)
         self.train_ids = self.dc.train_ids
--- a/src/extra/components/detector_motors.py
+++ b/src/extra/components/detector_motors.py
@@ -70,29 +70,33 @@ class DetectorMotors:
 
             self.keys.append(self.dc[src, key])
 
-    def positions(self, labelled=True):
+    def positions(self, labelled=True, compressed=False):
         """Returns the motor positions for all trains.
 
         Args:
             labelled (bool):
                 If True, returns the xarray with labelled dimensions,
                 overwise returns numpy.ndarray
-
+            compressed (bool):
+                If True, returns positions only when they change,
+                overwise positions in all recorded trains
         Returns:
-            positions (numpy.ndarray or xarray.DataArray):
+            positions (tuple of two numpy.ndarray or xarray.DataArray):
                 The motor positions
         """
         train_ids, pos, counts = self._read_positions()
-        pos = np.repeat(pos, counts, axis=0)
+        if not compressed:
+            pos = np.repeat(pos, counts, axis=0)
+            train_ids = self.train_ids
 
         if labelled:
             dims = ["trainId"] + list(self.coordinates.keys())
-            coords = {"trainId": self.train_ids}
+            coords = {"trainId": train_ids}
             coords.update(self.coordinates)
             return xarray.DataArray(
                 pos, dims=dims, coords=coords, name=self._position_key)
         else:
-            return pos
+            return train_ids, pos
 
     def _read_positions(self):
         """Reads and compresses motor positions."""
@@ -118,28 +122,6 @@ class DetectorMotors:
 
         self._positions = train_ids, pos[ix], counts
         return self._positions
-
-    def compressed_positions(self, labelled=True):
-        """Returns the unique motor positions and corresponding train IDs.
-
-        Args:
-            labelled (bool):
-                If True, returns the xarray with labelled dimensions,
-                overwise returns tuple of numpy.ndarrays
-
-        Returns:
-            positions (tuple of two numpy.ndarray or xarray.DataArray):
-                The motor positions
-        """
-        train_ids, values, _ = self._read_positions()
-        if labelled:
-            dims = ["trainId"] + list(self.coordinates.keys())
-            coords = {"trainId": train_ids}
-            coords.update(self.coordinates)
-            return xarray.DataArray(
-                values, dims=dims, coords=coords, name=self._position_key)
-        else:
-            return train_ids, values
 
     def positions_at(self, tid):
         """Returns motor positions at given train.
--- a/src/extra/components/detector_motors.py
+++ b/src/extra/components/detector_motors.py
@@ -147,18 +147,6 @@ class DetectorMotors:
         _, values, counts = self._read_positions()
         return values[np.argmax(counts)]
 
-    @property
-    def first(self):
-        """Returns the first motor positions."""
-        tid = self.train_ids[0]
-        return self.positions_at(tid)
-
-    @property
-    def last(self):
-        """Returns the last motor positions."""
-        tid = self.train_ids[-1]
-        return self.positions_at(tid)
-
     def __repr__(self):
         if not hasattr(self, "_ts"):
             self._ts = self.dc.train_timestamps()[0]
--- a/src/extra/components/detector_motors.py
+++ b/src/extra/components/detector_motors.py
@@ -408,6 +408,7 @@ class AGIPD1MQuadrantMotors(DetectorMotors):
             pattern = pattern.format(detector_id=detector_id, q="{q}", m="{m}")
             device_ptrn, key_ptrn = find_motors(
                 dc, pattern, self._position_key, q=groups, m=motors)
+            device_id = detector_id
 
         super().__init__(dc, device_id, device_ptrn, key_ptrn,
                          q=groups, m=motors)
--- a/src/extra/components/detector_motors.py
+++ b/src/extra/components/detector_motors.py
@@ -1,12 +1,19 @@
+import random
+import re
+import string
+from itertools import product
+
 import numpy as np
 import xarray
 
 
 def mangle_device_id_underscore(device_id):
+    """Converts deviceId to property notation (underscore version)."""
     return ''.join(c if c.isalnum() else '_' for c in device_id)
 
 
 def mangle_device_id_camelcase(device_id):
+    """Converts deviceId to property notation (camelcase version)."""
     # Replace non alpha-numeric chars with blanks
     mangled_id = ''.join(c if c.isalnum() else ' ' for c in device_id)
     # Make camelCase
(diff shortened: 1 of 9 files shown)


takluyver:
I think this broadly looks good, but I'd suggest we call it something more specific like `AGIPD1MQuadrantMotors` instead of the generic `DetectorMotors`. I don't know any specific details, but my guess is that if we add another detector type, it will be neater to do that with a separate class, even if we factor out a shared base class.

I'd like at least one of @philsmt or @JamesWrigley to take a look as well, with a particular focus on:

- Are there any patterns in other components this should follow for consistency?
- Is the 'data selector' logic likely to be useful elsewhere, and should it be factored out somehow?

JamesWrigley:
I apologize for procrastinating this for so long :see_no_evil: I'll take a squiz at it tomorrow.

egorsobolev:
> I think this broadly looks good, but I'd suggest we call it something more specific like `AGIPD1MQuadrantMotors` instead of the generic `DetectorMotors`. I don't know any specific details, but my guess is that if we add another detector type, it will be neater to do that with a separate class, even if we factor out a shared base class.

It seems that DSSC quadrant motors follow the same structure:
```python
    "SQS_NQS_DSSC": {
        "num_groups": 4,
        "num_motors": 2,
        "motor_pattern": "{detector_id}/MOTOR/Q{q}M{m}",
        "data_selector": None,
    },
```

takluyver:
Are we confident enough in that to put those definitions in? (Also asking anyone else who's familiar with DSSC, I guess @philsmt might know)

src/extra/components/detector_motors.py
```diff
@@ -0,0 +1,282 @@
...
+        "MID_EXP_AGIPD1M": {
+            "num_groups": 4,
+            "num_motors": 2,
+            "motor_pattern": "{detector_id}/MOTOR/Q{q}M{m}",
+            "data_selector": "MID_EXP_AGIPD1M1/MDL/DATA_SELECTOR",
+        },
+    }
+    ARGS = ("num_groups", "num_motors", "motor_pattern", "data_selector")
+
+    def __init__(self, dc, detector_id="", groups=None, motors_per_group=None,
```
takluyver, line 65: If only `dc` is passed, do we want to check for the detector IDs we know about, and if exactly one of them is found, use that? I think this is a common pattern for components, that `Component(run)` tries to find the relevant sources if it can.

(This is also a point in favour of making detector-specific subclasses, because it lets you be more specific without having to copy the device ID)

src/extra/components/detector_motors.py
```diff
@@ -0,0 +1,282 @@
...
+        detector_param = self.DETECTORS.get(detector_id, {})
+        for key in self.ARGS:
+            value = locals().get(key)
+            value = detector_param.get(key, value)
+            setattr(self, key, value)
+
+        args = (self.num_groups, self.num_motors, self.motor_pattern)
+        if any(arg is None for arg in args):
+            raise ValueError(
+                "You need specify all parameters for custom detector")
```
JamesWrigley, line 95: Nitpicking:
```suggestion
                "You need to specify all parameters for custom detector")
```

src/extra/components/detector_motors.py
```diff
@@ -0,0 +1,282 @@
...
+                for i in range(self.num_sources)
+            ]
+        return self._motor_ids
+
+    @property
+    def train_ids(self):
+        """The list of train IDs."""
+        return self.dc.train_ids
+
+    def train_id_coordinates(self):
```
JamesWrigley, line 154: I don't think it's necessary to have a property for the train IDs, but if we do we should just have one that returns either a list or array. Extra-data uses lists for historical reasons but I don't think it matters for us.

src/extra/components/detector_motors.py
```diff
@@ -0,0 +1,282 @@
...
+        """The list of motor device IDs."""
+        if not hasattr(self, "_motor_ids"):
+            self._motor_ids = [
+                self.motor_pattern.format(
+                    detector_id=self.detector_id,
+                    q=i // self.num_motors + 1,
+                    m=i % self.num_motors + 1,
+                )
+                for i in range(self.num_sources)
+            ]
```
JamesWrigley, line 146: Since this is pretty cheap to compute I think it's a bit cleaner to put this in `__init__()`, same for the other places where the `hasattr(...)` pattern is used.

JamesWrigley (changes requested):
Out of curiosity, do you anticipate taking data where the quadrant positions change during the run? Asking because there's quite a few methods in the component to give it a 'train-oriented' API and I haven't encountered a need for that yet. Nothing wrong with it of course, I'm just curious :)

This should also have some basic tests and docs (and a changelog entry).

JamesWrigley:
>     * Is the 'data selector' logic likely to be useful elsewhere, and should it be factored out somehow?

I don't think it's necessary right now, there are other 'systems' at MID for which motors are also saved in data selectors (e.g. the sample hexapod and focusing lenses) but I don't foresee making components for them anytime soon.

egorsobolev:
> Out of curiosity, do you anticipate taking data where the quadrant positions change during the run? Asking because there's quite a few methods in the component to give it a 'train-oriented' API and I haven't encountered a need for that yet. Nothing wrong with it of course, I'm just curious :)

You are right, SPB also avoid moving detector during acquisition. But once we forgot to stop a run and end up with changes during the run. There is also an option to union runs in extra-data. Since the geometry update is cheap, why not to have train-resolved data just in case?

> This should also have some basic tests and docs (and a changelog entry).
Sure, thanks. Let me do it after we decide on interface

src/extra/components/detector_motors.py
egorsobolev: This one for sure, thanks. But others methods read the data. This is not necessary cheap. From another side, if we instantiate the component, it implies reading. Should it read in constructor? I think it is not expected in general...

src/extra/components/detector_motors.py
takluyver: I kind of like the consistency of having `.train_ids` (list of selected trains) and `.train_id_coordinates()` (array aligned with data you can load). Certainly where those are different, I think that's a good idea. I don't have a strong preference where they're guaranteed to be the same, though.

JamesWrigley: `train_id_coordinates()` is indeed the closest to what this does in extra-data, but semantically I don't think it makes much sense in this context so my preference would be for `train_ids`. But I don't feel very strongly about it.

src/extra/components/detector_motors.py
JamesWrigley: Yeah you're right, NVM about the others then.

src/extra/components/detector_motors.py
egorsobolev: I have removed `.train_id_coordinates()`. Since motor positions are slow data, data coordinates looks less related.

src/extra/components/detector_motors.py
egorsobolev: I've generalized the base class a bit and created child class for AGIPD1M quadrants. In the last one I have implemented the search of the any suitable motors as sources or within any `DataSelector`. The data access logic remains in the base class.

Is it better now?

src/extra/components/detector_motors.py
```diff
@@ -0,0 +1,414 @@
...
+            tid (int):
+                Train ID
+
+        Returns:
+            postions (numpy.ndarray):
+                The motor positions
+        """
+        train_ids, values, _ = self._read_positions()
+        i = np.searchsorted(train_ids, tid, side="right")
+        return values[max(i - 1, 0)]
```
JamesWrigley, line 184: I think this should throw an error if the `tid` is not in `self.train_ids`.

src/extra/components/detector_motors.py
```diff
@@ -0,0 +1,414 @@
...
+        ix = np.flatnonzero(
+            np.insert(np.any(np.diff(pos, axis=0), axis=axes), 0, True))
+
+        # train ids when positions are changed
+        train_ids = np.asarray(self.dc.train_ids)[ix]
+        # count of trains at every position
+        counts = np.diff(np.append(train_ids.astype(int),
+                                   int(self.dc.train_ids[-1]) + 1))
+
+        self._positions = train_ids, pos[ix], counts
```
JamesWrigley, line 146: I dunno about saving the compressed arrays as an internal representation, it feels unnecessarily complicated. Why not store the entire position arrays in memory?

JamesWrigley (changes requested):

src/extra/components/detector_motors.py
egorsobolev: Good catch!

src/extra/components/detector_motors.py
egorsobolev: I'm not sure if DAQ creates records for all trains. It should, in principle, but I've decided to not rely on it for protection reason. With compressed version we aren't attached to the stored trains anymore, and can return motor positions for any train between changes.

Compressed version is also allows to see directly if there are changes during the run.

Since this, it makes sense to expose compressed version. If we expose, then why not to store as compressed? The compression and decompression aren't really complicated. They are just four and one line.

JamesWrigley: Ehhh, we definitely shouldn't make up data for trains that aren't saved :sweat_smile: But if `positions_at()` will throw when given a bad train ID then I guess it won't happen anyway.

> Since this, it makes sense to expose compressed version. If we expose, then why not to store as compressed? The compression and decompression aren't really complicated. They are just four and one line.

So one reason would be a simpler internal representation, the other would be a more efficient implementation of `positions_at()` because it wouldn't have to do a search every time (though TBF it's presumably a binary search so it won't cost much anyway). But I don't feel strongly about it so let's continue with it as is.

src/extra/components/detector_motors.py
```diff
@@ -0,0 +1,414 @@
...
+        return self._positions
+
+    def compressed_positions(self, labelled=True):
```
JamesWrigley, line 149: This is rather nitpicky, but I don't really like the generic term `labelled`. Both the EXtra components and extra-data functions use that name to return either `Series` or `DataArray` objects, which feels confusing/inconsistent to me :shrug:

I'd prefer a more explicit `xarray` argument, though I know this would add even more inconsistency so I don't insist on it.

src/extra/components/detector_motors.py
```diff
@@ -0,0 +1,414 @@
...
+            if key.endswith(suffix))
+        src_ptrn = pattern_camelcase.format(**frm_args)
+        num_sources = count_sources(keys, src_ptrn, **dims)
+        if num_sources == num_motors:
+            return data_selector_id, src_ptrn + '.' + position_key
+
+        src_ptrn = pattern_underscore.format(**frm_args)
+        num_sources = count_sources(keys, src_ptrn, **dims)
+        if num_sources == num_motors:
+            return data_selector_id, src_ptrn + '.' + position_key
```
takluyver, line 353: This is a big chunk of pretty complicated code to identify sources & keys. I guess the aim is to generalise so it will work for other detectors in the future, but I don't think it's worth it - I think writing the code specifically for AGIPD and then doing something similar for other detectors if/when we need it will end up being easier to understand.

src/extra/components/detector_motors.py
```diff
@@ -0,0 +1,414 @@
...
+    def first(self):
+        """Returns the first motor positions."""
+        tid = self.train_ids[0]
+        return self.positions_at(tid)
+
+    @property
+    def last(self):
+        """Returns the last motor positions."""
+        tid = self.train_ids[-1]
+        return self.positions_at(tid)
```
takluyver, line 202: I would make these regular methods, rather than properties, because calling them may need to load data. This is a judgment call, but generally properties (which you access like `.first` without an explicit call) are expected to be things that are cheap to compute.

If I see methods like `.first()`, or even `.positions_at()`, I'd also assume that these load only the required bit of data, so they're faster than loading all the data and indexing it. But these work by loading all the data, which seems a bit surprising.

src/extra/components/detector_motors.py
```diff
@@ -0,0 +1,414 @@
...
+                src_ptrn.format(**dict(zip(names, values))),
+                key_ptrn.format(**dict(zip(names, values)))
+            ) for values in product(*coordinates)]
+        self.motor_labels = [
+            ''.join(f"{n.upper()}{v}" for n, v in zip(names, values))
+            for values in product(*coordinates)]
+
+        self.shape = tuple(len(c) for c in coordinates)
+        self.num_sources = np.prod(self.shape)
+        self.dc = dc.select(self.motor_ids)
```
takluyver, line 86:
```suggestion
        self.dc = dc.select(self.motor_ids, require_all=True)
```

In general this shouldn't make a difference for control sources, but with multiple data aggregators writing a run it's possible that some of them start or finish a train before the others.

src/extra/components/detector_motors.py
takluyver, line 149: I wonder if something like `unique_positions` would be a clearer name than 'compressed'?

src/extra/components/detector_motors.py
egorsobolev: I agree, it is over complicated. I think I could considerably simplify it, keeping still general. But I have implemented searching only for known or specified detector_id

src/extra/components/detector_motors.py
egorsobolev: Good point, thanks!

src/extra/components/detector_motors.py
egorsobolev: I agree, but I tend to follow the extra-data naming until we agree another convention.

egorsobolev: I don't know. The problem that positions may be not unique here. I decided to combine `compressed_positions` and `positions` by adding the compressed argument in the last.

src/extra/components/detector_motors.py
egorsobolev: I would really surprised if the loading of one number is really much faster than load entire vector. Of course there could be a corner cases.

I will remove first and last at all. I don't really think that they are useful in the interface. For the `position_at`, I think it is reasonable compromise to read a vector once.

egorsobolev:
Sorry, it took again long time - operations... Thanks for your suggestions, I think I have addressed all of them. Please take a look

src/extra/components/detector_motors.py
```diff
@@ -0,0 +1,265 @@
...
+
+        Args:
+            tid (int):
+                Train ID
+
+        Returns:
+            postions (numpy.ndarray):
+                The motor positions
+        """
+        if tid < self.train_ids[0] or self.train_ids[-1] < tid:
```
JamesWrigley, line 137: I think this is more correct:
```suggestion
        if tid not in self.train_ids:
```

Otherwise this would still return a value if `tid` happens to be a missing train.

JamesWrigley (changes requested):

src/extra/components/detector_motors.py
egorsobolev: True, it is... if we want to accurately, reliably 100% protect.

From other side, it is too expensive (even more expensive than protected code), taking into account that actually should never happen (DAQ populates all trains during aggregation)

If it happens anyway, can we do the same as DAQ  does - populate the missed value with previous? I think this is reasonable compromise, and users/instruments would prefer this instead of dropping image. At least for lit-frame selection, I had to guess of missed data.

I think we can be not so strict here and let users implement more strict protection by themself when it's necessary (or filter data collection properly, e.g. `select` with `require_all=True`)

What do you think?

takluyver: > this actually should never happen (DAQ populates all trains for control data during aggregation)

I believe the DAQ has the possibility to pause recording a run and then resume it later. We can also pool the data from two separate runs (with `.union()`), which will likely have a gap between them. So it's unusual, but definitely not impossible to have missing trains in the middle of control data.

JamesWrigley: I would prefer to be conservative, I don't think our components should ever give data for a train that isn't present. And it seems unlikely that this function will ever be in a hot path so I'd say the performance hit is ok.

philsmt: I would also say that we should never make contiguity assumptions about train IDs for `CONTROL` data, and treat them as arbitrary sets of numbers.

While it is true that it is unlikely to happen for the actual files, it's not uncommon to apply `.select(..., require_all=True)` across all sources early, causing gaps in `CONTROL` data due to gaps in `INSTRUMENT` data, e.g.:

```python
subrun = run.select('*AGIPD*DET*:xtdf', '*MOTOR*')
AGIPD1MQuadrantMotors(subrun)
```

If you're concerned about performance, I wonder if `np.searchsorted` would be faster than `np.isin(unique=True)`. You can use the assumption that train IDs are sorted and unique.

egorsobolev: In all these cases, there should not be images as well, right?

JamesWrigley: To me that's not a convincing argument :sweat_smile:

egorsobolev: `searchsorted` wins, if `trains_ids` pre-cast to `ndarray`.

image

egorsobolev: Ok, the `tid` protection is strict now. please take a look and let's finish it!

egorsobolev:
Thanks all for review!

PR #155, opened 2024-04-07
Some minor improvements to generic pulse components

philsmt:
This PR collates a couple of minor changes by themselves, aimed at building on the earlier https://github.com/European-XFEL/EXtra/pull/138. The commits mostly build on top of each other, but I tried to keep them clean by focusing on a single aspect.

* 214c635a4ca87db5e451fb964ee64465fbda7ac8 Adds an undocumented method of constructing `TimeserverPulses` components by directly passing a `SourceData` object. This is used internally to construct different `PulsePattern`-based components after the fact from the same data without having its `DataCollection`.
* 3505f247241fa9e169ee35ce8c781662706ff8f9 Using the feature above, this adds a method `.machine_pulses()` to easily obtain a `MachinePulses` component from any other timeserver-related component. As this will be often used just to get the machine repetition rate, there's an even more explicit method for that.
* a8cf3a34147c2585be2e0f4c73b513c07a7b6ec5 Per request from users, this starts adding @lru_cache to some methods where applicable. Also added a warning that `PulsePattern` are expected to be immutable.
* 21142a3e3e846fd660cdae3cfebe980aaebea601 adds a generic method to detect interleaving pulse patterns, and an explicit version to conveniently check this for SA1 and SA3.

### DIFF
--- a/src/extra/components/pulses.py
+++ b/src/extra/components/pulses.py
@@ -333,6 +333,58 @@ class PulsePattern:
             return True
 
 
+    def is_interleaved_with(self, other: PulsePattern) -> Optional[bool]:
+        """Check whether pulses are interleaved with other pattern.
+
+        This method returns True if and only if it can safely determine
+        that the pulse pattern described by this component and the
+        passed component are interleaved, and False if and only if it is
+        clear they are not interleaved. In all other cases, the pulses
+        may or may not be interleaved and None is returned.
+
+        Returns:
+            (bool or None) Whether pulses are interleaved or None if the
+                pulse patterns do not allow the determination.
+        """
+
+        for (_, self_pids), (_, other_pids) in zip(
+            self.pulse_ids(copy=False).groupby(level=0),
+            other.pulse_ids(copy=False).groupby(level=0)
+        ):
+            self_num = len(self_pids)
+            other_num = len(other_pids)
+
+            if self_num == 0 or other_num == 0:
+                # No statement can be drawn from missing pulses.
+                continue
+
+            self_first_pulse = self_pids.min()
+            other_first_pulse = other_pids.min()
+
+            if self_num > 1 and other_num > 1:
+                # Multiple pulses in both self and other.
+
+                if self_first_pulse < other_first_pulse:
+                    # self early, interleaved <=> other starts within self
+                    return self_pids.max() > other_first_pulse
+                else:
+                    # other early, interleaved <=> self starts within other
+                    return other_pids.max() > self_first_pulse
+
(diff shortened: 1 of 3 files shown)


src/extra/components/pulses.py
```diff
@@ -290,6 +295,7 @@ def get_pulse_mask(self, *args, **kwargs):
              DeprecationWarning, stacklevel=2)
         return self.pulse_mask(*args, **kwargs)
 
+    @lru_cache
```
JamesWrigley, line 298: I think we can get away with using [`@cache`](https://docs.python.org/3/library/functools.html#functools.cache) here since the return value will never change.

src/extra/components/pulses.py
```diff
@@ -310,6 +317,58 @@ def is_constant_pattern(self):
             all([len(x) == 1 for x in pulse_ids.groupby(level=1).unique()])
         )
 
+    def is_interleaved_with(self, other: PulsePattern) -> Optional[bool]:
+        """Check whether pulses are interleaved with other pattern.
+
+        This method returns True if and only if it can be safely
+        determined the pulse pattern described by this component and the
```
JamesWrigley, line 324: Nitpicking:
```suggestion
        This method returns `True` if and only if it can safely
        determine that the pulse pattern described by this component and the
```

src/extra/components/pulses.py
```diff
@@ -806,6 +866,52 @@ def bunch_pattern_table(self) -> KeyData:
...
         return self._key
 
+    def machine_pulses(self, **kwargs) -> MachinePulses:
+        """Get MachinePulses component for the same data.
+
+        Any additional keyword arguments are passed to the MachinePulses
+        initializer.
+
+        Returns:
+            (MachinePulses) Correspondig component using the same data
```
JamesWrigley, line 876: Nitpicking:
```suggestion
            (MachinePulses) Corresponding component using the same data
```

JamesWrigley (changes requested):

src/extra/components/pulses.py
philsmt: Thanks!

src/extra/components/pulses.py
philsmt: Thanks!

src/extra/components/pulses.py
JamesWrigley: Alrighty, makes sense.

JamesWrigley (approved):
LGTM!

philsmt:
Thanks for review!

Issue #136, opened 2024-03-06
PulsePattern.is_constant_pattern ignores empty trains


philsmt:
It's even mentioned in the documentation, but this seems kind of useless.

JamesWrigley:
Closing this because #156 is merged now.

PR #156, opened 2024-04-07
Take empty trains into account when checking for constant pulse pattern

philsmt:
As requested via https://github.com/European-XFEL/EXtra/issues/136 (actually from instrument staff), this PR adds an optional flag to `PulsePattern.is_constant_pattern()` to optionally take empty trains into account when determining whether the pulse pattern is content. An empty train will essentially always the pattern to not be constant. To preserve the original API, they continue to be ignored by default.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -17,6 +17,7 @@
     ```
 Added:
 
+- Empty trains can now optionally be included when determining constant pulse patterns via [XrayPulses.is_constant_pattern()][extra.components.XrayPulses.is_constant_pattern] (!156).
 - A helper function named [fit_gaussian()][extra.utils.fit_gaussian] (!131).
 
 Fixed:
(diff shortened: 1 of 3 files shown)


JamesWrigley (approved):
LGTM!

tmichela:
Minor suggestion: I would make the added argument keyword only, for readability.

Else LGTM.

philsmt:
> Minor suggestion: I would make the added argument keyword only, for readability.

Fair point 🤔 

I think I also want to invert the logic, make it `False` by default and *turn it to `True`* as an option.

philsmt:
Made the argument keyword-only and inverted the logic, will merge tomorrow afternoon if there's no objection.

JamesWrigley:
The change is fine, but it needs a `Breaking` entry in the changelog.

philsmt:
Why? 
I'm sorry if this wasn't clear, but I meant inverting the logic of the new keyword while keeping the previous default behaviour 🙂

JamesWrigley:
Oh right, sorry I misunderstood, NVM then.

PR #161, opened 2024-04-16
Improve handling of XGM data recording the wrong number of pulses

JamesWrigley:
In particular, by:
- Explicitly checking for a wrong number of pulses by comparing fast/slow data and emitting a warning if they differ.
- Add more obvious warnings to the docs.

Follow up to #153.


src/extra/components/xgm.py
```diff
@@ -391,19 +394,21 @@ def npulses(self, sase=None) -> int:
...
         return self._npulses[pg]
 
     def pulse_counts(self, sase=None):
         """Return a 1D [DataArray][xarray.DataArray] of the number of pulses in each train.
 
-        See the docs for [XGM.npulses()][extra.components.XGM.npulses]
-        for information on retrieving the true number of pulses.
+        Warning:
+            This can be unreliable, see the docs for
+            [XGM.npulses()][extra.components.XGM.npulses] for information on
```
takluyver, line 410: For someone in a hurry to get something done, and maybe not a native English speaker, I think it would be easy to read this as 'this method is bad, `npulses()` is better'.

I think it might be worth repeating the pointer to `XrayPulses` in each of these warnings, even if that's a bit more verbose, rather than pointing people to the docs of another method to find that information.

src/extra/components/xgm.py
```diff
@@ -417,13 +422,24 @@ def pulse_counts(self, sase=None):
...
             self._npulses_by_train[pg] = self._control_source[key].xarray()
 
+            slow_counts = self._npulses_by_train[pg]
+            pulse_energy = self.pulse_energy(sase=sase)
+            common_tids = np.intersect1d(pulse_energy.trainId, slow_counts.trainId)
+            fast_counts = np.count_nonzero(pulse_energy.sel(trainId=common_tids), axis=1)
+
+            counts_match = np.allclose(fast_counts, slow_counts.sel(trainId=common_tids))
+            if not counts_match:
+                warn(f"Slow data pulse counts ({key}) don't match the counts from fast data (data.intensityTD), data may be invalid!")
```
takluyver, line 432: If we know the slow data counts aren't always reliable and the difference from the fast data is a useful signal of that, does that imply the fast data are more reliable? Should we be using that instead?

JamesWrigley: Yes, in my experience the fast data is more reliable and in some ways more useful than what we get from the pulses components because it will show the real number of pulses that are getting delivered. I don't know why but sometimes the number of pulses will change for... reasons... and for whatever reason that doesn't seem to be reflected in the bunch pattern table.

As for whether we should be using that instead of the slow data, I'm not sure. I am a little hesitant to make it do something other than the obvious 'just return whatever was saved', but as you say that's probably not very useful.

philsmt: > it will show the real number of pulses that are getting delivered. I don't know why but sometimes the number of pulses will change for... reasons... and for whatever reason that doesn't seem to be reflected in the bunch pattern table.

Hu, do you have a concrete example? That should not be possible, as they both are looking at exactly the same data - the XGM DOOCS server takes the bunch pattern to slice out the correct values from the train signal for `data.intensityTD`. That's the entire reason I don't trust it, it's an interpretation of the same signal through a 3rd party I don't control.

JamesWrigley: Ok, so weirdly enough the `XrayPulses` component does report the right number but not `PumpProbePulses` :octopus: Example from p6156, r185 with the XGM:
image

And the pulses components:
image

My code was using `PumpProbePulses` so it was giving the wrong number of pulses. Is that a bug in `PumpProbePulses`?

philsmt: No, it's doing exactly what it is supposed to do:

image


`PumpProbePulses` creates a unified pattern from both, i.e. it looks at every possible pulse the machine could have, and counts it as a filled pulse if **either** FEL, PPL **or both** are present.

JamesWrigley: Ah ok, makes sense.

philsmt (approved):
LGTM.

We could have actually offered a `PulsePattern` object based on XGM data, this would have automatically given all the different methods like constant pattern, pulse counts etc.

src/extra/components/xgm.py
```diff
@@ -391,19 +394,21 @@ def npulses(self, sase=None) -> int:
...
+            pulse_counts = self.pulse_counts(sase=sase)
+            if not np.allclose(pulse_counts[0], pulse_counts):
+                raise ValueError("Number of pulses is changing, there is no nominal number.")
+
+            self._npulses[pg] = int(pulse_counts[0])
 
         return self._npulses[pg]
 
     def pulse_counts(self, sase=None):
         """Return a 1D [DataArray][xarray.DataArray] of the number of pulses in each train.
```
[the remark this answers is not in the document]
JamesWrigley: No reason, it's just a preference since I tend to use DataArrays/Datasets a lot.

JamesWrigley:
I think I'm happy with the state of this now, in 53d15d4 I changed the `pulse_counts()` to return the fast data counts by default if a mismatch was detected. I'll leave this open for another round of review, but feel free to rebase and merge it if everyone agrees.

src/extra/components/xgm.py
JamesWrigley: Sorry, I realize I never replied to this. The warnings are now removed from all those methods because `.pulse_counts()` will fall back to the fast data if there are mismatches and emit a warning.

PR #162, opened 2024-04-16
[XGM] Add support for .pulse_energy() returning a Series, and .slow_train_energy()

### LINKED ISSUE #141: Make a method to retrieve slow pulse energy data with XGM component
XGM component can retrieve only pulse-resolved pulse energy data at the moment. Sometimes we also use train-resolved (slow) data. It usually has a key of a form 'controlData.slowTrainSa<x>'. Please add a method for retrieving it.

### LINKED ISSUE #142: Add a method to XGM component that retrieves fast pulse energy data as Series
At the moment there's a method to get the pulse energy as a 2D array truncated at the largest number of pulses/train number in the run. It would be good also to have this data in a form of series with only relevant entries. Please add a method for this (or a flag argument to the existing method specifying the form of the output).

JamesWrigley:
I also added pandas to our dependencies since it wasn't explicitly in there already.

Fixes #141, fixes #142 (CC @ser-u1).

### DIFF
--- a/src/extra/components/xgm.py
+++ b/src/extra/components/xgm.py
@@ -299,14 +299,19 @@ class XGM:
 
         return PropertyGroup(sase) if sase is not None else self._default_pg
 
-    def pulse_energy(self, sase=None):
-        """Return a 2D [DataArray][xarray.DataArray] of the pulse energy in microjoules.
+    def pulse_energy(self, sase=None, series=False):
+        """Returns the energy per-pulse and per-train in microjoules.
 
-        This has dimensions of `(trainId, pulseIndex)`. For runs with a varying
-        number of pulses, the data will be sliced to the *maximum number* of
-        pulses. e.g. if a run has 100 trains with only one train containing 10
-        pulses and all the others 0, the returned array will have a shape of
-        `(100, 10)`.
+        If `series=False` (the default) this will return a 2D
+        [DataArray][xarray.DataArray] with dimensions of `(trainId,
+        pulseIndex)`. For runs with a varying number of pulses, the data will be
+        sliced to the *maximum number* of pulses. e.g. if a run has 100 trains
+        with only one train containing 10 pulses and all the others 0, the
+        returned array will have a shape of `(100, 10)`.
+
+        If `series=True` this will return a 1D [Series][pandas.Series] where all
+        entries with 0 pulses have been dropped, indexed by `trainId` and
+        `pulseIndex`.
 
         Note:
             This uses the `data.intensityTD` property of the XGM.
@@ -319,6 +324,8 @@ class XGM:
                 `data.intensitySa1TD` with `sase=1`. This setting overrides the
                 `default_sase` argument to the
                 [constructor][extra.components.XGM].
+            series (bool): Whether to return a 2D [DataArray][xarray.DataArray]
+                or 1D [Series][pandas.Series].
         """
         pg = self._check_sase_arg(sase)
         if pg not in self._pulse_energy:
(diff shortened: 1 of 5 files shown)


src/extra/components/xgm.py
```diff
@@ -353,7 +361,36 @@ def pulse_energy(self, sase=None):
 
             self._pulse_energy[pg] = pulse_energy
 
-        return self._pulse_energy[pg]
+        pulse_energy = self._pulse_energy[pg]
+        if series:
+            # Use .where() to convert 0's to nans, then convert to a Series,
+            # then drop the nan entries.
+            pulse_energy = pulse_energy.where(pulse_energy > 0).to_series().dropna()
```
tmichela, line 368: units are lost at that point, you need to add them back.

tmichela (approved):
Just minor comments, else LGTM

src/extra/components/xgm.py
```diff
@@ -353,7 +361,36 @@ def pulse_energy(self, sase=None):
...
+                [XGM.pulse_energy()][extra.components.XGM.pulse_energy].
+        """
+        pg = self._check_sase_arg(sase)
+        if pg not in self._slow_train_energy:
+            if pg == PropertyGroup.MAIN:
+                key = "controlData.slowTrain"
+            else:
+                key = f"controlData.slowTrainSa{pg.value}"
+
+            energy = self.control_source[key].xarray()
```
philsmt, line 388: For scalar values, `pd.Series` may be preferred here as well, though of course it is less critical on a train boundary. It should be cheap to offer though.

tests/test_components_xgm.py
```diff
@@ -58,6 +59,8 @@ def test_xgm_getters(multi_xgm_run, source_sase):
     xgm.photon_energy_by_train()
     xgm.doocs_server()
     xgm.pulse_energy()
+    xgm.slow_train_energy()
+    assert isinstance(xgm.pulse_energy(series=True), pd.Series)
```
philsmt, line 63: We're not checking the actual data? 😇

src/extra/components/xgm.py
```diff
@@ -319,6 +325,8 @@ def pulse_energy(self, sase=None):
                 `data.intensitySa1TD` with `sase=1`. This setting overrides the
                 `default_sase` argument to the
                 [constructor][extra.components.XGM].
+            series (bool): Wherther to return a 2D [DataArray][xarray.DataArray]
```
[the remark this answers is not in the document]
philsmt: 👍

philsmt (approved):
Sigh, I really don't trust the pulse selection done by the XGM, when comparing over larger runs there are regularly mismatches between it and timeserver data. But alas, it's easy to access.

Some minor nitpicks, otherwise LGTM.

src/extra/components/xgm.py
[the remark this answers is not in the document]
JamesWrigley: I disagree for two reasons:
- Personally I dislike it when some vague term is used that implies completely different behaviour, like returning a `Series` instead of a `DataArray`.
- `flatten` is ambiguous because it could also mean returning a 1D `DataArray` with a `pd.MultiIndex`, rather than a `Series`.

An alternative would be `flatten_as_series`, though that's a bit long for my taste.

src/extra/components/xgm.py
```diff
@@ -353,7 +361,36 @@ def pulse_energy(self, sase=None):
...
+        """
+        pg = self._check_sase_arg(sase)
+        if pg not in self._slow_train_energy:
+            if pg == PropertyGroup.MAIN:
+                key = "controlData.slowTrain"
+            else:
+                key = f"controlData.slowTrainSa{pg.value}"
+
+            energy = self.control_source[key].xarray()
+            energy.attrs["units"] = "µJ"
```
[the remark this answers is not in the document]
JamesWrigley: Fixed in 4ca92a9.

src/extra/components/xgm.py
JamesWrigley: Fixed in 3ad2174.

tests/test_components_xgm.py
JamesWrigley: I didn't think it was necessary since we're basically just calling `DataArray.to_series()`, but I added tests for it anyway in 3ad2174.

philsmt: Thanks. I just try to not rely on implementation but interface for what needs testing 🙂

src/extra/components/xgm.py
philsmt: By chance I came across a run with negative XGM values. I dislike magic values for filtering, but at least we shoudl go for `== 0`.

JamesWrigley: Fixed in 8bed29a.

src/extra/components/xgm.py
philsmt: That's fine, no rush. We should discuss how we can try to make the types we return predictable in the long run. This componet is now out as it is anyway.

PR #164, opened 2024-04-24
Expose the DAMNIT API

JamesWrigley:
So that we can do `from extra.damnit import Damnit`.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -17,6 +17,7 @@
     ```
 Added:
 
+- A module to expose the [DAMNIT API](damnit.md).
 - [AdqRawChannel][extra.components.AdqRawChannel] to access and process raw traces saved by fast ADQ digitizers (!35).
 - Hits obtained via [DelayLineDetector.hits()][extra.components.DelayLineDetector.hits] can now optionally be restricted to hits reconstructed with a certain method. (!170).
 - [DldPulses][extra.components.DldPulses] can optionally enumerate PPL-only pulses using negative indices to stay compatible with PPL-unaware data. (!167).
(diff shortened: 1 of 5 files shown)


takluyver:
The implementation LGTM.

I have some nebulous doubts about the broader pattern, though. What do we gain by giving people an option to write `from extra.damnit` instead of `from damnit`? We're not making one big unified package, because we're still versioning & documenting the pieces independently. We could argue that it makes the `EXtra` docs a central pointer to all the different pieces, but we've got the [user docs](https://rtd.xfel.eu/docs/data-analysis-user-documentation/en/latest/) as a central place (& we should ensure DAMNIT is there, of course), and having two separate central places isn't great.

Is exposing packages like this a stepping stone towards actually bringing together code in one big EXtra package? I could see we *might* do that for `.data` and `.geom` (not convinced we should). But it seems like more of a stretch either to bring DAMNIT into EXtra or to break DAMNIT's code up to move part of it here. :thinking:

JamesWrigley:
> We're not making one big unified package, because we're still versioning & documenting the pieces independently.

We're not making one big package, but we are absolutely making a single entrypoint :stuck_out_tongue: To quote the introduction:
> EXtra is a library developed at the European XFEL to provide a single entrypoint for data analysis to users.

The DA docs is a separate beast to me because it also covers things like Jupyter and environments, whereas the extra docs is specifically about libraries for accessing/analyzing the data. I think I would be fine with eventually replacing the `Software` page (or at least most of it) with a link to extra.

takluyver:
Sorry, I meant to say sooner, feel free to carry on. I still have some doubts about the utility of a single entry point, but that doesn't need to hold this up.

PR #166, opened 2024-04-29
Add support to align pulse-indexed data as new columns in DLD sparse data

philsmt:
Adds a new keyword argument to `dld.signals()` and `dld.hits()` to take data indexed by pulse, align it with the sparse DLD data and add as a column to the signals/hits dataframe. This allows something like:

```python3
delays = OpticalLaserDelay(run).total_delays()
hits = DelayLineDetector(run).hits(extra_columns={'delay': delays})
```

This could in principle also be added to `dld.edges()` with minor modifications, but less likely to be useful at the moment.

### DIFF
--- a/src/extra/components/dld.py
+++ b/src/extra/components/dld.py
@@ -168,6 +168,31 @@ class DelayLineDetector:
 
         return pd_cls(raw[finite_mask], pd.MultiIndex.from_frame(index_df))
 
+    def _insert_aligned_columns(self, df, columns):
+        """Add pulse-indexed data to reduced dataframe.
+
+        Args:
+            df (pandas.DataFrame): Frame to insert columns into.
+            columns (dict): Mapping of column name to pulse-indexed data
+                to insert. Must be re-indexable by internal pulse index
+                and data is repeated for multiple entries in a pulse.
+
+        Returns:
+            None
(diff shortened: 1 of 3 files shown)


src/extra/components/dld.py
```diff
@@ -269,32 +294,50 @@ def edges(self, channel_index=True, pulse_dim='pulseId'):
...
+    def signals(self, pulse_dim='pulseId', extra_columns={}):
         """Get reconstructed signals as dataframe.
 
         Args:
             pulse_dim ({pulseId, pulseIndex, time}, optional): Label
                 for pulse dimension, pulse ID by default.
+            extra_columns (dict, optional): Mapping of column name to
+                pulse-indexed data to align and insert into dataframe,
+                must be re-indexable to internal pulse index. Data is
+                repeated for multiple entries of a single pulse.
```
takluyver, line 306: I think the columns have to be pandas Series, right? It might be worth saying this explicitly.

Series objects can also have a name attribute, so we could consider allowing a list and using `s.name` as the column name. But that can be done later.

takluyver:
Other than the suggestion about docs, LGTM

philsmt:
Thanks for the quick review!

PR #167, opened 2024-05-01
Optionally use negative pulse indices for PPL-only DLD pulses

philsmt:
The `PumpProbePulses` component and its spiritual predecessor `DldPulses` combine FEL-only, PPL-only and pumped pulses into a single consistent pattern, i.e. may have more pulses than a pure FEL pattern in the presence of PPL-only pulses.

While this is useful for analysis, it presents problems when correlating FEL-only data using enumerated indices to label pulses. The XGM component e.g. will only ever look at pulses with FEL (either FEL-only or pumped) and enumerate accordingly, being incompatible with the integrated pattern.

While this can be fixed by using global pulse IDs (i.e. referring to indices in the machine pattern table), the use of pulse indices is still often desired and/or easier to obtain. This PR adds an optional flag to `DldPulses` (only for now, can be ported to `PumpProbePulses`) to modify the enumeration behaviour:

* FEL-only and pumped pulses are enumerated without taking the PPL into account
* PPL-only pulses are _enumerated_ with negative indices, i.e. starting with -1 and counting down

The resulting indices are again compatible with PPL-unaware pulse indices, while still containing the PPL-only pulses. 

Examples:

* Four FEL pulses with two preceding PPL-only pulses:
  `0, 1, 2, 3, 4, 5` becomes `-1, -2, 0, 1, 2, 3`
* Four FEL pulses with two PPL-only pulses at start and end each
  `0, 1, 2, 3, 4, 5, 6, 7` becomes `-1, -2, 0, 1, 2, 3, -3, -4`

Also fixes a minor bug with the preceding https://github.com/European-XFEL/EXtra/pull/166 to ensure indexes are compatible, and simplifies passing custom arguments to construct the internal pulse information with `DelayLineDetector`.

### DIFF
--- a/src/extra/components/pulses.py
+++ b/src/extra/components/pulses.py
@@ -1465,25 +1465,56 @@ class DldPulses(PulsePattern):
         first_pulse_id (int, optional): Pulse ID for the first pulse,
             only used in case of missing pulse ID information in data
             and 0 by default.
+        negative_ppl_indices (bool, optional): Whether to enumerate
+            PPL-only pulses with negative indices (while keeping
+            non-negative indices for FEL-only or pumped pulses) or
+            interleave all pulses with positive indices (default).
     """
 
-    def __init__(self, detector, *, clock_ratio=None, first_pulse_id=None):
+    def __init__(self, detector, *, clock_ratio=None, first_pulse_id=None,
+                 negative_ppl_indices=False):
         super().__init__(detector, detector['raw.triggers'])
 
         self._clock_ratio = clock_ratio
         self._first_pulse_id = first_pulse_id
+        self._negative_ppl_indices = negative_ppl_indices
 
     def _get_train_ids(self):
         return np.unique(self._key.train_id_coordinates())
 
     def _get_pulse_ids(self):
         triggers = self._key.ndarray()
+        train_ids = self._key.train_id_coordinates()
+
+        if self._negative_ppl_indices:
+            if 'ppl' not in triggers.dtype.fields:
+                raise ValueError('negative PPL indices only available '
+                                 'with PPL information')
+
+            pulse_indices = np.zeros(len(triggers), dtype=int)
+            prev_tid = -1
+
+            for i, tid, row in zip(range(len(triggers)), train_ids, triggers):
+                if tid != prev_tid:
+                    prev_tid = tid
+                    only_ppl_index = -1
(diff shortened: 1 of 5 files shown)


takluyver:
When I first glanced at the title, I was thinking of a scenario where all your pump-only pulses come before the first X-ray pulse. In that specific case, it would be natural to use negative indices like:

```
[-4, -3, -2, -1, 0, 1, 2, 3]
```

Obviously that doesn't work neatly in other scenarios, like the one in your tests. But if this scenario does occur, I wonder if it will be confusing to people that the order is:

```
[-1, -2, -3, -4, 0, 1, 2, 3]
```

I don't think there's anything we can do about that, but I thought I'd mention it in case it gives you a great idea. :wink:

src/extra/components/pulses.py
```diff
@@ -1465,25 +1465,56 @@ class DldPulses(PulsePattern):
...
+                                 'with PPL information')
+
+            pulse_indices = np.zeros(len(triggers), dtype=int)
+            prev_tid = -1
+
+            for i, tid, row in zip(range(len(triggers)), train_ids, triggers):
+                if tid != prev_tid:
+                    prev_tid = tid
+                    only_ppl_index = -1
+                    with_fel_index = 0
```
takluyver, line 1501: There's probably a way to do this neatly with [`itertools.groupby()`](https://docs.python.org/3/library/itertools.html#itertools.groupby) to have an outer loop over train IDs. It's also not a big deal, though - I know you want to get this out quickly, and I think this code works fine.

takluyver:
Other than that, LGTM

src/extra/components/pulses.py
philsmt: I tried

```python
for train_id, vals in groupby(enumerate(triggers), lambda x: train_ids[x[0]]):
    only_ppl_index = -1
    with_fel_index = 0

    for i, trigger in vals:
        if trigger['ppl'] and not trigger['fel']:
            pulse_indices[i] = only_ppl_index
            only_ppl_index -= 1
        else:
            pulse_indices[i] = with_fel_index
            with_fel_index += 1
```

but it made the overall call to `pulse_ids()` about 15% slower. While I suspect the grouping itself is faster, the nested loops seem to end up taking more time due to constructing the necessary generators.

PR #169, opened 2024-05-02
Add Scan.split_by_steps() method

takluyver:
To split any of our objects with a `.select_trains()` method, as well as suitable DataArrays. E.g. we could have used this recently for FXE to split JUNGFRAU data into scan steps before loading it.

This is pretty trivial to do without an extra method, but it seems like a nice convenience, and I think it fits neatly with `bin_by_steps`. I also hope that we can later optimise it to avoid the N-squared performance we had with `.split_trains()` on long runs in the past.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -27,6 +27,7 @@ Added:
 - Check whether any pulse patterns are interleaved with `is_interleaved_with()` or directly SA1/SA3 with `is_sa1_interleaved_with_sa3()` (!155).
 - Obtain [MachinePulses][extra.components.MachinePulses] from any other timeserver-based pulse components directly via `machine_pulses()` or get machine repetition rate directly from `machine_repetition_rate()` (!155).
 - A helper function named [fit_gaussian()][extra.utils.fit_gaussian] (!131).
+- A new method [Scan.split_by_steps()][extra.components.Scan.split_by_steps] (!169).
 
 Fixed:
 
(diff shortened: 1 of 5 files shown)


philsmt:
`TrainSelectable` protocol 👀 

LGTM!

philsmt (approved):

takluyver:
Thanks!

Do you know if the generated data for tests includes any scan-like data to test with? I added to one of the existing tests, but it's a bit useless since there's no real data, so zero scan steps to work with.

philsmt:
Unfortunately I'm not aware of any. In EXtra-data I added data in individual tests, while in EXtra there's only what you can see in `mockdata`. FWIW, I will be in need of a `Motor` with actual data for the `OpticalLaserDelay` tests as well.

takluyver:
I've added some simple scan-like data for `MOTOR/MCMOTORYFACE.actualPosition`, to test this better.

JamesWrigley:
FWIW the tests already use `Scan._mkscan()` to generate fake scans: https://github.com/European-XFEL/EXtra/blob/6f03c90fd903c5913db7234dd6ec0e68a9416182/src/extra/components/scan.py#L308

takluyver:
Thanks! I half spotted that, but I wanted to make the test a bit more realistic, with data coming from files, and to have the train IDs line up with the rest of the run for splitting it.

PR #170, opened 2024-05-06
Add support to filter DLD hits for reconstruction method while loading

philsmt:
The DLD reconstruction performed as part of the calibration pipeline assigns a reconstruction method to each hit, an enum value denoting which signals were successfully paired to reconstruct the hit ranging from 0 (perfect) to 19 (mostly guessing). This is well documented in the processing reports with various plots.

While the dataframe loaded via `DelayLineDetector.hits` can easily be loaded filtered manually, it is very common to only consider unrisky reconstructions during regular analysis. Also, in some scenarios there can be many badly reconstructed hits with only a few good ones, making all following operations slower even if those hits are not interesting.

This PR adds support to filter by reconstruction method early while masking the input data for finite entries. Any subsequent indexing operation will therefore only operate on the desired rows.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -17,7 +17,8 @@
     ```
 Added:
 
-- [DldPulses](extra.components.DldPulses) can optionally enumerate PPL-only pulses using negative indices to stay compatible with PPL-unaware data.
+- Hits obtained via [DelayLineDetector.hits()][extra.components.DelayLineDetector.hits] can now optionally be restricted to hits reconstructed with a certain method. (!170)
+- [DldPulses](extra.components.DldPulses) can optionally enumerate PPL-only pulses using negative indices to stay compatible with PPL-unaware data. (!167)
 - The `slowTrains` XGM property is available through
   [XGM.slow_train_energy()][extra.components.XGM.slow_train_energy] (!162).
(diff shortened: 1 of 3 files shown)


philsmt:
Hmm good point, let me get back to you on that. Indeed I set the limit to `None` initially to not break the API, but the idea came out of frustration while using it. 

Yes, it's that DLD. You're right the docstring actually never went that far in explicitly referring to the offline processing we're offering based on raw detector data. By now there are fully proprietary detectors with black boxes in operation that directly present some x,y,t data to Karabo and are saved.

philsmt:
Frankly I haven't looked much into the other DLD's data yet, but yes, it's also some form of sparse data saved in fixed length arrays with padding. David actually already inquired about a component for that.

philsmt:
After discussions, we indeed decided to go for filtering non-risky hits by default (14). I added an explicit comment in the docstring for awareness. Generally risky hits were already ignored in the past after the first few analysis steps, as they are rarely useful.

I also clarified the docstring for the entire module to indicate it's about data from offline reconstructions, not proprietary detectors that immediately save such data for the DAQ.

takluyver:
LGTM

philsmt:
Thanks for review

PR #172, opened 2024-05-07
Add bad pixel flags to extra.calibration

takluyver:
The definitions are copied from pycalibration.

### DIFF
--- a/docs/calibration.md
+++ b/docs/calibration.md
@@ -55,8 +55,12 @@ agipd_cd = CalibrationData.from_report(3757)
 
 The calibration pipeline produces masks along with corrected data, in keys
 called `image.mask` or `data.mask` depending on the detector. Zeros in the mask
-represent good pixels, while any other value indicates one or more reasons to
-consider the corresponding data dubious:
+represent normal, good pixels, while any other value indicates one or more
+reasons why the data may be atypical.
+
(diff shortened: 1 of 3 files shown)


docs/calibration.md
```diff
@@ -50,3 +50,14 @@ agipd_cd = CalibrationData.from_report(3757)
 ::: extra.calibration.JUNGFRAUConditions
 
 ::: extra.calibration.LPDConditions
+
+## Bad pixel values
+
+The calibration pipeline produces masks along with corrected data, in keys
+called `image.mask` or `data.mask` depending on the detector. Zeros in the mask
+represent good pixels, while any other value indicates one or more reasons to
+consider the corresponding data dubious:
```
philsmt, line 59: I would prefer as neutral wording as possible, given some bits simply mean to inform about a _peculiarity_ like `NON_STANDARD_SIZE`. Maybe conspicuous? (just grabbing that from a dictionary)

takluyver: Good point, I've had a go at rewording this.

Are you happy with the subheading remaining 'Bad pixel values' if we give more nuance in the description,  or do you think that needs to change too?

philsmt: I suspect that term is just too established to change now 🤷

PR #173, opened 2024-05-13
Support more types for DLD dataframe extra columns

philsmt:
Building on https://github.com/European-XFEL/EXtra/pull/166, this PR relaxes the type requirements a bit to work with both pulse- and train-resolved data in `pd.Series`, `xr.DataArray` or directly `KeyData` objects. Naturally the pulse index must be *compatible*, i.e. use the same first two columns.

### DIFF
--- a/src/extra/components/dld.py
+++ b/src/extra/components/dld.py
@@ -5,6 +5,9 @@ import re
 import numpy as np
 import pandas as pd
 
+from extra.data import KeyData
+from .utils import _isinstance_no_import
+
 
 class DelayLineDetector:
     """Interface for processed delay line detector data.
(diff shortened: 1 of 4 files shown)


src/extra/components/dld.py
```diff
@@ -169,32 +172,64 @@ def _build_reduced_pd(self, data, index, entry_level):
 
         return pd_cls(raw[finite_mask], pd.MultiIndex.from_frame(index_df))
 
-    def _insert_aligned_columns(self, df, columns):
+    @staticmethod
+    def insert_aligned_columns(df, columns):
```
takluyver, line 176: Is it a deliberate choice to make this method public? I think it will also appear in the docs, and I tend to think that keeping the documented API relatively small, so people can easily see the methods they want, is a good thing. But YMMV. :slightly_smiling_face:

takluyver:
The implementation looks good to me.

It doesn't seem 100% ideal to be making APIs that look like `load_some_data(with=some_other_data)`, as opposed to some form of `load_some_data() + some_other_data`. But I think that's hard to avoid given the constraints of what you're doing here, so go ahead.

philsmt:
Thanks for review!

Indeed I still want a `PulsePattern.label_data`-esque API that can take data in as many different formats as possible and align it. But it seemed unlikely to get this right quickly, so I opted for a smaller integration in the DLD component itself. It should be easy to continue supporting this API on top of a more generic one later.

takluyver:
In that case, LGTM

PR #174, opened 2024-05-13
Add parameter to get only FEL or only pump laser mask from PumpProbePulses

takluyver:
This should allow easier ways to get information like 'which FEL pulses have a corresponding optical laser pulse':

```python
ppl_on = ppp.select_trains(0).pulse_mask(field='ppl', labelled=False).squeeze()
fel_on = ppp.select_trains(0).pulse_mask(field='fel', labelled=False).squeeze()

fel_pulse_is_pumped = ppl_on[fel_on]
```

(I also explored another option in the branch [`pumpprobe-dataframe`](https://github.com/European-XFEL/EXtra/compare/pumpprobe-dataframe); I think this variant fits in better with the existing API, with less extra stuff to learn.)

### DIFF
--- a/src/extra/components/pulses.py
+++ b/src/extra/components/pulses.py
@@ -281,9 +281,6 @@ class PulsePattern:
 
         Returns:
             (xarray.DataArray or numpy.ndarray):
-
-        Returns:
-            (numpy.ndarray or pandas.Series):
         """
 
         mask = self._get_pulse_mask()
@@ -1445,6 +1442,16 @@ class PumpProbePulses(XrayPulses, OpticalLaserPulses):
         table but only contains boolean flags whether a given pulse
         was present in this pattern.
 
+        For instance, to see whether each FEL pulse has a corresponding pump
+        laser pulse in a consistent pulse pattern:
+
+        ```python
+        ppl_on = ppp.select_trains(0).pulse_mask(field='ppl').squeeze()
+        fel_on = ppp.select_trains(0).pulse_mask(field='fel').squeeze()
+
+        fel_pulse_is_pumped = ppl_on[fel_on]
+        ```
+
         Args:
             labelled (bool, optional): Whether a labelled xarray
                 DataArray (default) or unlabelled numpy array is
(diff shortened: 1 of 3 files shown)


src/extra/components/pulses.py
```diff
@@ -1438,10 +1438,36 @@ def _get_pulse_mask(self, reduced=False):
...
+            labelled (bool, optional): Whether a labelled xarray
+                DataArray (default) or unlabelled numpy array is
+                returned.
+            field (str, optional): 'fel'/'ppl' to mark only FEL or only pump
+                laser pulses in the returned array. By default, pulses are True
+                whether either an FEL pulse or a PPL pulse occured.
+
+        Returns:
+            (xarray.DataArray or numpy.ndarray):
+
```
philsmt, line 1458: But what does it return now? 😁

philsmt (approved):
Thanks, that sounds like a useful addition. LGTM beyond the two doc-related comments.

src/extra/components/pulses.py
takluyver: Oops, copypasta. The method on the parent class also has two `Returns:`. I'll fix both.

src/extra/components/pulses.py
```diff
@@ -1438,10 +1438,36 @@ def _get_pulse_mask(self, reduced=False):
...
-    @wraps(PulsePattern.pulse_mask)
-    def pulse_mask(self, labelled=True):
-        return TimeserverPulses.pulse_mask(
-            self, labelled=labelled).astype(bool)
+    def pulse_mask(self, labelled=True, field=None):
+        """Get boolean pulse mask.
+
+        The returned mask has the same shape as the full bunch pattern
+        table but only contains boolean flags whether a given pulse
+        was present in this pattern.
```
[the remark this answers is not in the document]
takluyver: I've added the example in.

JamesWrigley:
I'm a bit late to the party here, but I just tried to use this and two things popped out at me:
- The ordering of the arguments has the `field` after `labelled`. Naively I would expect people to mostly want a specific field as shown in the example, so would it make sense to make that the first argument so people don't have to specify it as a keyword argument?
- It's very weird to me that we keep using `labelled` to arbitrarily refer to either a `Series` or `DataArray` :stuck_out_tongue: And it's doubly weird here because the rest of the pulses components exclusively use `labelled` to mean `Series`. What do we think about replacing/deprecating it with explicit `series`/`xarray` arguments instead?

PR #177, opened 2024-05-22
Expose extra_data components in extra.components

takluyver:
This would avoid people needing to know that some component classes belong to EXtra and others to EXtra-data, if they can import all of them from `extra.components`.

On the other hand, it could create more confusion if someone ends up with an older EXtra-data package and a newer EXtra, or vice versa, because these classes appear to be part of EXtra, but actually aren't.

I don't have a strong preference, but I figured a PR was a good way to have the discussion.


JamesWrigley (approved):
I lean towards agreeing with this, but we should mention it in the documentation.

philsmt (approved):
Can we merge this?
There didn't seem to be strong feelings either way, and this *works*

JamesWrigley:
I still think it needs docs but I can add that later.

PR #178, opened 2024-05-24
Deprecate pulse dimension time in favor of pulseTime

philsmt:
In hindsight, optionally using the name `time` in the pulse index was kind of bound for name collisions. I would therefore rather sooner than later deprecate it in favor of `pulseTime`. This MR will continue to allow `time` to work, but no longer documents it and raises a warning if used.


JamesWrigley (approved):
Out of curiosity, what was the conflict with `time`?

philsmt:
I would like to offer `samples` and `time` as coordinates in the digitizer component, too 🫣

PR #189, opened 2024-06-07
Fix insertion of aligned columns when original frame has none

philsmt:
129a07902a193d3cde1e9a49ba141fddb99e13c2 This slightly improves the behaviour of `DelayLineDetector.insert_aligned_columns` in the case the dataframe the columns are inserted to does not actually have any columns.

While this can never be the case in `DelayLineDetector` itself, this method is likely become more generically used over time 
(and as it happens is already used in the wild by @bj-s, who also suggested this fix) 

117e7b38a5715880880dc93fbdab58e5d7286bb7 Some additional docstring fixes

### DIFF
--- a/src/extra/components/dld.py
+++ b/src/extra/components/dld.py
@@ -230,16 +230,15 @@ class DelayLineDetector:
             if shared_index[:2] == df.index.names[:2]:
                 # Same pulse dimensions as the dataframe.
                 if num_per_pulse is None:
-                    num_per_pulse = df[df.columns[0]].groupby(
-                        level=df.index.names[:-1]).count()
+                    num_per_pulse = df.groupby(
+                        level=df.index.names[:-1]).size()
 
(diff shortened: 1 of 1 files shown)


daniloefl:
It looks good.

One general comment: I like that you put the type in the docstrings, but why not use the `x: int` or `x: Optional[ArrayLike]` etc to indicate the type with `typing`? Probably not for this PR, though.

philsmt:
Thanks!

> One general comment: I like that you put the type in the docstrings, but why not use the `x: int` or `x: Optional[ArrayLike]` etc to indicate the type with `typing`? Probably not for this PR, though.

Mostly because we're not in a place yet where we'd make much use of `typing` annotations, and for that I dislike the visual clutter it tends to introduce. I'm not against it in general, and it's likely where the ecosystem is headed. But at the moment, the benefit is fairly limited.

PR #190, opened 2024-06-07
Docs fixes, and document `PulsePattern.select_trains()`

JamesWrigley:
I was initially just documenting `PulsePattern.select_trains()` but then I saw some other issues so I fixed them along the way. That removes all warnings for me from mkdocs.

### DIFF
--- a/src/extra/components/dld.py
+++ b/src/extra/components/dld.py
@@ -201,7 +201,7 @@ class DelayLineDetector:
                 index and data is repeated accordingly.
 
         Returns:
-            None
+            (NoneType): None
         """
 
         # Compute these lazily.
--- a/.gitignore
+++ b/.gitignore
@@ -99,3 +99,6 @@ ENV/
 *.h5
 
 src/extra/__version__.py
+
+# Generated C files from Cython
+*.c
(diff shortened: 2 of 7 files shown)


src/extra/components/dld.py
```diff
@@ -357,7 +357,7 @@ def signals(self, pulse_dim='pulseId', extra_columns={}, max_method=None):
         This data is primarily for detector diagnostics purposes and
         should generally not be used for scientific data analysis,
         please refer instead to reconstructed x, y, t data obtained from
-        [hits()](extra.components.DelayLineDetector.hits()).
```
philsmt, line 360: Could you please rebase? 🙈 

I fixed these issues specifically in 117e7b38a5715880880dc93fbdab58e5d7286bb7 earlier.

philsmt:
Rebase should remove some changes already done, all other LGTM!

Thanks, it's high time we are polishing docs a bit. I hope I get next to some examples using p700000.

src/extra/components/dld.py
JamesWrigley: Ah yes you're right, I missed that... I've rebased it now.

PR #191, opened 2024-06-10
Two line docstring fix for DelayLineDetector

TimBerberich:
Small docstring correction related to commit 203cf34.


philsmt:
Thanks, well spotted. They indeed produce a [different result](https://extra.readthedocs.io/en/latest/components/delay-line-detectors/#extra.components.DelayLineDetector.signals) in the docs.

LGTM.

PR #193, opened 2024-06-13
fix: convert conditions for CALCAT to float or string as expected.

kakhahmed:
This Pull request is connected to https://git.xfel.eu/calibration/pycalibration/-/merge_requests/1019

CALCAT converts conditions values to float if digit and to string otherwise. This achieves the same goal so the application code solves the issue in case a boolean condition is given. Currently, it will be sent as a string, and a wrong value is used in CALCAT but with PR it will be converted to floats.

### DIFF
--- a/src/extra/calibration.py
+++ b/src/extra/calibration.py
@@ -525,9 +525,19 @@ class CalibrationData(Mapping):
             (dict) Operating condition for use in CalCat API.
         """
 
+        def to_float_or_string(value):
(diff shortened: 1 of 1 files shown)


takluyver:
Thanks! Can you add a note to `docs/changelog.md` to briefly summarise what this fixes? Otherwise LGTM.

kakhahmed:
I have updated the `changelog`  👍🏽

takluyver:
Thanks!

Issue #185, opened 2024-05-31
Add tests for AdqChannel


philsmt:

takluyver:
There's data we could use in XMPL, e.g. p700002 r8. The source is `FXE_RR_DAQ/ADC/1:network`. Channel 2C is used for I0 measurements, if having actual signal is helpful for writing the tests. The value dips during a pulse:

![Image](https://github.com/user-attachments/assets/2590042b-1998-4544-a721-1b6f944a5a76)

PR #196, opened 2024-06-25
Fix using a labelled series to compute pulse distances

philsmt:
Before someone states the obvious, this is why one should have unit tests.

Turns out substracting two pandas series with different indices from each other gives unexpected results (`{nan, 0.0}` rather than a finite number).


JamesWrigley (approved):
I assume this falls under #185 :stuck_out_tongue: If you don't think it's worth a changelog entry you can add the `skip-changelog` label. Anyway, LGTM!

PR #197, opened 2024-06-26
Add missing EdgeInterpolation docs and various other text fixes

philsmt:
I missed including the `EdgeInterpolation` enum in the docs, and took the opportunity to expand a bit more on it as well as fix some other miscellaneous docs in the same module.

### DIFF
--- a/src/extra/signal/_ftd.pyx
+++ b/src/extra/signal/_ftd.pyx
@@ -19,13 +19,49 @@ import numpy as np
 ctypedef floating data_t
 ctypedef floating[::contiguous] data_array_t
 
+
 cpdef enum EdgeInterpolation:
-    # Interpolation method to find edge position.
+    """Edge interpolation methods.
+
+    The fast timing discriminators in this module can use several
+    different interpolation methods to estimate the real edge position
+    in between the actual samples. These differ in terms of their
+    interpolation quality and performance requirements.
+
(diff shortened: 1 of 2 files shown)


src/extra/signal/_ftd.pyx
```diff
@@ -19,14 +19,51 @@ import numpy as np
...
+    Some of the interpolation methods can be configured via
+    [config_ftd_interpolation][extra.signal.config_ftd_interpolation].
+
+    Attributes:
+        NEAREST: Chose the sample position whose value is closest to the
+            value at the true edge position, no cost in performance and
+            will always return an integer result.
+        LINEAR: Perform a linear interpolation between the two bordering
+            samples to find the position.
+        SPLINE: NYI.
```
JamesWrigley, line 54: Nitpicking, I'd use `Not yet implemented` in full here to avoid confusion.

JamesWrigley (changes requested):

src/extra/signal/_ftd.pyx
philsmt: You're right. On second thought though, I'm not sure why include it in the first place. It's an enum anyway, so we can add it later any time.

JamesWrigley (approved):
LGTM!

philsmt:
Thanks!

PR #198, opened 2024-06-26
Fix extra_compile_args in setup.py

philsmt:
Frankly I have no idea if this is a legacy key, or I simply messed things up when I added more Cython extensions. But it should definitely be `extra_compile_args`.


takluyver:
This isn't something I know all that much about, but LGTM

philsmt:
Thanks!

For reference: https://setuptools.pypa.io/en/latest/userguide/ext_modules.html#setuptools.Extension

PR #201, opened 2024-07-03
Attempt to improve plots of scan positions

takluyver:
I don't find the current `Scan.plot()` plots super informative - steps are distinguished by matplotlib's default colour cycle, but the colours aren't all easily distinguishable on a thin line. So here's my attempt: I've dropped the different colours, but drawn rectangles instead using the motor resolution to make it clearer how the steps are distinguished. And I've labelled a few of the steps so it's easier to find specific steps in big scans.

Here it is on a scan with just a few steps:

image

And a scan with a lot of steps:

image

And zooming in on that:

image


takluyver:
Better colour suggestions welcome - the line is currently matplotlib's default blue, and I've randomly picked a pink shade for the resolution boxes.

JamesWrigley (changes requested):
I think it looks good overall, could you update the screenshot in the docs too?

And what does it look like with bad results, like if the resolution or some other parameter is way off?

takluyver:
Good question. Here it is with the resolution set too high, making it clear that something is wrong:

image

Setting `min_trains=1` so it picks up the 'tail' on this scan isn't very obvious, unless you know you were expecting 41 steps:

image

Is it worth doing something to make this case more obvious? The best idea I've got right now is to draw some additional marker, maybe an x or an o, at the start or centre of each scan step.

FXE scans always seem to get very neat motor positions, so I don't have an easy way to test it on noisy data.

takluyver:
Perhaps a more realistic illustration of the resolution going wrong: here it's set to the step size, which is 0.05. This is about double the resolution it guesses if you don't pass it (which behaves nicely).

image

takluyver:
Using centre markers to indicate tiny steps more clearly on the plot (from a `Scan(..., min_trains=1)`:

image

image

If this is only likely to affect the last step, like in this example, another option would be to ensure that's always labelled. But I don't particularly like that this won't catch tiny steps elsewhere.

image

JamesWrigley:
I think that second option with the crosses looks the cleanest. What does it look like without the centre markers and with the old color labelling?

takluyver:
Good question, this is what it looks like with the current plot method:

image

JamesWrigley:
Ok, I'd say let's go for the one with the crosses :+1:

JamesWrigley (approved):
LGTM!

takluyver:
Thanks James!

PR #202, opened 2024-07-04
Markdown table methods for calibration constants

takluyver:
Copied over from the internal calibration repository.

The saved HTTP responses ('cassettes') needed refreshing, because they were previously created before `view_url` was added to the response. I updated them all in case we want to write other tests against this.


philsmt (approved):
Thanks, LGTM! We should soon be able to use EXtra directly.

Only commend if you have the time, it may be nice to use the table in the docs example.

takluyver:
Is this what you had in mind?

https://extra--202.org.readthedocs.build/en/202/calibration/

philsmt:
Yes! Just seems like the kind of feature that's especially nice for an example.

PR #207, opened 2024-07-22
Add Python 3.12 to CI

JamesWrigley:
I don't think we have any plans to use Python 3.12 soon, but 3.12 seems to be surprisingly breaking so I'd like to start testing on it early.


takluyver:
LGTM

PR #208, opened 2024-07-26
Prepare changelog for new release 2024.1.2 

tmichela:


philsmt:
I think we used to do this everytime for a semesterly release, so LGTM.

docs/changelog.md
```diff
@@ -5,16 +5,18 @@
...
+<!--     in a personal environment you'll need to install the package from git. -->
 
-    ```bash title="Installation command"
-    pip install git+https://github.com/European-XFEL/EXtra.git
-    ```
+<!--     ```bash title="Installation command" -->
+<!--     pip install git+https://github.com/European-XFEL/EXtra.git -->
+<!--     ``` -->
+
+## [2024.2]
```
JamesWrigley, line 19: This needs to have a link to the tag at the bottom of the file (doesn't matter if the tag doesn't exist yet).

JamesWrigley (changes requested):
In hindsight it's a bit confusing to tag releases at the beginning of a new cycle rather than at the end :thinking: Because then someone wanting to install extra for an old experiment in cycle `i` will need to install version `year.{i + 1}` (possibly wrapping around the year) to get any changes that were introduced in the cycle `i`. Should we change to tagging them at the end instead?

tmichela:
That makes sense indeed. Should we then tag that one `2024.1.2` and switch to `2024.2` at the end of this cycle?

JamesWrigley:
Sure, sounds good :+1:

JamesWrigley (approved):
LGTM!

PR #210, opened 2024-07-30
feat: add new begin_at_strategy argument and split dataset_obj()

kakhahmed:
- Expose `begin_at_stragety` calcat argument to switch between prior and closest strategies
- Split dataset_obj() deriving path part to different function, as it can be usefull to use externally

This MR is related to an already almost available change in pycalibration:https://git.xfel.eu/calibration/pycalibration/-/merge_requests/1035/

This is related to that issue as well : https://git.xfel.eu/calibration/planning/-/issues/200 of adding `begin_at_strategy` to `calibration_constant_versions/get_by_detector_conditions` API


kakhahmed:
This MR is ready to review from my side. The coverage tests are a bit rough on me :(

I see that there are no `SingleConstant` test methods, therefore I didnt try adding tests for the new `file_path` method

PR #212, opened 2024-07-31
Add support for Karabacon 3.0.10

JamesWrigley:
This is somewhat speculative because I haven't tested it on real data yet. See: https://git.xfel.eu/karaboDevices/Karabacon/-/merge_requests/460

The property type changed this time so more changes are likely necessary.


tmichela:
LGTM

PR #215, opened 2024-08-02
Add infrastructure and initial set of tests for ADQ component

philsmt:
Adds the infrastructure (mock device and sources) to test ADQ digitizers and a basic set of tests for initialization and some auxiliary methods of the ADQ component.

I rather split this up, as the next methods will require significantly more complex tests.

### DIFF
--- a/src/extra/components/adq.py
+++ b/src/extra/components/adq.py
@@ -494,7 +494,7 @@ class AdqRawChannel:
     def channel_key(self, suffix):
         """Instrument KeyData object of this channel."""
         return self._instrument_src[f'digitizers.channel_{self._channel_name}'
-                                    f'_{suffix}']
+                                    f'.{suffix}']
 
     @property
     def raw_samples_key(self):
--- a/src/extra/components/adq.py
+++ b/src/extra/components/adq.py
@@ -124,7 +124,7 @@ class AdqRawChannel:
             raise ValueError('channel expected to be 2 or 3 characters, '
                              'e.g. 1A or 1_A')
 
-        self._channel_number = ord(self._channel_letter) - ord('A')
+        self._channel_number = ord(self._channel_letter) - ord('A') + 1
         self._channel_name = f'{self._channel_board}_{self._channel_letter}'
 
         key = f'digitizers.channel_{self._channel_name}.raw.samples'
--- a/src/extra/components/adq.py
+++ b/src/extra/components/adq.py
@@ -232,6 +232,10 @@ class AdqRawChannel:
         if baselevel is not None:
             baseline = baseline - baselevel
 
+        # Make sure the dtypes match, otherwise baseline is likely
+        # going to be float64 and not castable via `safe`.
+        baseline = baseline.astype(out.dtype, copy=False)
+
         for offset in range(period):
             sel = np.s_[offset::period]
             np.subtract(
@@ -707,7 +711,8 @@ class AdqRawChannel:
                 baseline to 0.
 
         Returns:
-            out (numpy.ndarray): Corrected input data.
+            out (numpy.ndarray): Corrected input data, same dtype as
+                input data if floating otherwise `float32`.
         """
 
         if cm_period < 1:
@@ -716,7 +721,8 @@ class AdqRawChannel:
         if not isinstance(data, np.ndarray):
             data = np.asarray(data)
 
-        out = np.zeros_like(data, dtype=np.float32)
+        out = np.zeros_like(data, dtype=data.dtype \
+            if np.issubdtype(data.dtype, np.floating) else np.float32)
         self._correct_cm_by_train(data, out, cm_period, baseline, baselevel)
 
         return out
--- a/tests/mockdata/adq.py
+++ b/tests/mockdata/adq.py
@@ -1,0 +1,82 @@
+
+import numpy as np
+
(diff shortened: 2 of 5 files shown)


philsmt:
Also contains a couple of fixes that came up when writing the tests...

tests/mockdata/adq.py
```diff
@@ -0,0 +1,75 @@
...
+from extra_data.tests.mockdata.base import DeviceBase
+
+from extra.utils import gaussian
+
+
+class AdqDigitizer(DeviceBase):
+    # still needed?
+    control_keys = [
+        ('fakeScalar', 'u4', ()),
+    ]
```
takluyver, line 13: Is there nothing really in the CONTROL section?

tests/mockdata/adq.py
```diff
@@ -0,0 +1,75 @@
...
+        super().write_instrument(f)
+
+        root_grp = f[f'INSTRUMENT/{self.device_id}:network/digitizers']
+        x = np.arange(150000)
+
+        for i, ch_label in enumerate(self.channel_labels):
+            # Add a channel-dependent baseline shift and gaussian to
+            # each channel.
+            root_grp[f'channel_{ch_label}/raw/samples'][:] += -10 * (i+1) \
+                + gaussian(x, 0, i*80, i*1000, 50).astype(np.int16)
```
takluyver, line 75: Would it make sense to do this optionally for a selected single channel? We don't generally fill mock files with random data, and I imagine this is going to make the tests slower & more memory intensive, especially on systems using tmpfs (i.e. temporary files are created in RAM).

tests/mockdata/adq.py
philsmt: Currently nothing of interest or that is used. The interesting values are only in the `RUN` section, in particular after we recently started using property filtering to work around serious aggregator issues when saving all `CONTROL` values.

tests/mockdata/adq.py
philsmt: Sure, it shouldn't be a problem to limit actual data to a few select channels. But unlike with EXtra-data, I'd consider real data in the files essential to test what the components do - in particular if they contain transformations.

philsmt: 87a9fb337012e12966cbc991fbdc4425712d976f limits data generation to channels selected in the fixture, and I reduced the trace length to 50000 (which should be enough for any realistic data-based tests). That should keep the memory cost negligible while the fixture is only generated once per session.

tests/mockdata/adq.py
takluyver: I assume you've put `fakeScalar` here because something goes wrong if it's an empty list? Is that something we should fix, or not worth the effort?

src/extra/components/adq.py
```diff
@@ -124,7 +124,7 @@ def __init__(self, data, channel, digitizer=None, pulses=None,
             raise ValueError('channel expected to be 2 or 3 characters, '
                              'e.g. 1A or 1_A')
 
-        self._channel_number = ord(self._channel_letter) - ord('A')
+        self._channel_number = ord(self._channel_letter) - ord('A') + 1
```
takluyver, line 127: Are we confident no-one is already using channel numbers as they are now, counting from zero?

philsmt: Yes, the numbers and letters are enumated at initialization time by the `AdqDigitizer` device and cannot be configured. They're not even (always) in the physical order in the crate.

tests/mockdata/adq.py
philsmt: I documented here a while ago, but did not get back to it yet: https://github.com/European-XFEL/EXtra-data/issues/429

src/extra/components/adq.py
takluyver: Sorry, perhaps I'm misunderstanding. At the moment, if I create `adq = AdqDigitizer(run,  '1C')`, then `adq.number` will give me 2 (A=0, B=1, C=2), yes? With this change, it becomes 3.  Are you confident that no-one has started using those numbers as they are?

(We can still change it even if someone might be using it - I just want to think about if we need to announce the change somehow)

tests/mockdata/adq.py
takluyver: Aha, thanks, I missed that.

takluyver:
Other than the question about announcing the numbering change, this LGTM.

src/extra/components/adq.py
philsmt: Apologies, I understand now. So the earlier behaviour was a bug, `adq.number` was always meant to represent the board _number_ in the sense it was used in `CONTROL` parameters (yes, someone funny used letters for fast data and numbers for slow data). As a matter of fact, `.channel_parameters` would've given you the wrong values as of now.

I highly doubt this parameter was used already directly, but if you feel strongly about it I can add it as a bugfix.

takluyver: OK, thanks. Let's mention it in the changelog, though since we install master every day it doesn't make much difference.

philsmt:
Wrapping this up, I added a changelog notice for the fix/change in enumeration of channels.

@takluyver Unless you object I will merge this by the end of the day.

takluyver:
Go for it :+1:

PR #221, opened 2024-08-21
Fit a unnormalized Gaussian by default in `fit_gaussian()`

JamesWrigley:
Getting back a normalized amplitude is a bit unexpected when you're fitting physical data.

It's a bit annoying to work around this so I'll merge it later tonight if there's no objections. Technically it is a bit breaking but I'm willing to accept that since it's fairly early in the cycle and this is a decent improvement.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -17,8 +17,15 @@
     ```
 
 Fixed:
+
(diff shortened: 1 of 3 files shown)


takluyver:
Retrospective LGTM

PR #222, opened 2024-08-22
Add A_sign parameter for fit_gaussian()

takluyver:
Specify Asign=1 for a peak pointing upwards, or Asign=-1 for a peak pointing downwards. The default leaves it unconstrained, but the initial parameter guesses in that case are for an upwards peak, preserving the current behaviour.

Specifying bounds make `curve_fit` switch to a slower method, but how much slower seems to be very variable. In one test, I went from ~4ms per fit to ~40ms, but in another test, only from about 2.3 ms to 3.6 ms. The default still uses no bounds so it can use the faster algorithm.

### DIFF
--- a/src/extra/utils/functions.py
+++ b/src/extra/utils/functions.py
@@ -50,8 +50,9 @@ def fit_gaussian(ydata, xdata=None, p0=None, norm=False, Asign=0, **kwargs):
         p0 (list): A list of `[y0, A, μ, σ]` to match the arguments to
             [gaussian()][extra.utils.gaussian].
         norm (bool): Whether to fit a normalized or unnormalized Gaussian.
-        Asign (int): 1 for an upwards peak, -1 for downwards. 0 (default) allows
-            either, using a faster algorithm. Passing `bounds=` overrides this.
+        Asign (int): Sign of the amplitude (A) parameter for the Gaussian.
+            1 for an upwards peak, -1 for downwards. 0 (default) allows either,
+            using a faster algorithm. Passing `bounds=` overrides this.
         **kwargs (): All other keyword arguments will be passed to
             [curve_fit()][scipy.optimize.curve_fit].
     """
--- a/src/extra/utils/functions.py
+++ b/src/extra/utils/functions.py
@@ -25,7 +25,7 @@ def gaussian(x, y0, A, μ, σ, norm=True):
     return y0 + (A / norm_factor) * np.exp(-(x - μ)**2 / (2 * σ**2))
 
 
-def fit_gaussian(ydata, xdata=None, p0=None, norm=False, **kwargs):
+def fit_gaussian(ydata, xdata=None, p0=None, norm=False, Asign=0, **kwargs):
     """Fit a Gaussian to some data.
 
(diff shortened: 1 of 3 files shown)


takluyver:
Design question: is a string parameter like `orientation='up'` preferable to the +-1 thing? My guess was that specifying the sign is natural for scientists.

src/extra/utils/functions.py
```diff
@@ -50,6 +50,8 @@ def fit_gaussian(ydata, xdata=None, p0=None, norm=False, **kwargs):
         p0 (list): A list of `[y0, A, μ, σ]` to match the arguments to
             [gaussian()][extra.utils.gaussian].
         norm (bool): Whether to fit a normalized or unnormalized Gaussian.
+        Asign (int): 1 for an upwards peak, -1 for downwards. 0 (default) allows
+            either, using a faster algorithm. Passing `bounds=` overrides this.
```
JamesWrigley, line 54: 'Asign' sounds almost like an English word so I'd explicitly mention that it's the amplitude sign. And what do you mean by faster algorithm? Adding bounds should make the algorithm converge faster so I'd expect setting a specific sign to improve performance.

JamesWrigley (changes requested):

JamesWrigley:
> Design question: is a string parameter like `orientation='up'` preferable to the +-1 thing? My guess was that specifying the sign is natural for scientists.

Nah I think I'd go with $$\pm 1$$, like you say that's probably more natural for most folks.

src/extra/utils/functions.py
JamesWrigley: Oh, sorry I didn't read the second half of your original post :upside_down_face: Ok, that's really weird...

takluyver: I think the speedup with no bounds is weird mathematically but makes some sense computationally. `curve_fit` has different algorithms behind it, and the default one depends on whether you specify any bounds:

> With `method='lm'`, the algorithm uses the Levenberg-Marquardt algorithm through [leastsq](https://docs.scipy.org/doc/scipy-1.14.0/reference/generated/scipy.optimize.leastsq.html#scipy.optimize.leastsq). Note that this algorithm can only deal with unconstrained problems.

src/extra/utils/functions.py
```diff
@@ -25,7 +25,7 @@ def gaussian(x, y0, A, μ, σ, norm=True):
     return y0 + (A / norm_factor) * np.exp(-(x - μ)**2 / (2 * σ**2))
 
 
-def fit_gaussian(ydata, xdata=None, p0=None, norm=False, **kwargs):
+def fit_gaussian(ydata, xdata=None, p0=None, norm=False, Asign=0, **kwargs):
```
JamesWrigley, line 28: Nitpicking, could we rename this to `A_sign`? I slightly prefer having the underscore for consistency with other possible parameters like `y0_min` etc.

JamesWrigley (changes requested):

src/extra/utils/functions.py
takluyver: Sure, that makes sense.

JamesWrigley (approved):
LGTM!

takluyver:
Thanks!

PR #227, opened 2024-09-03
Fix sign for stage-to-time conversion in OpticalLaserDelay

philsmt:
This is a bit embarassing, and I'm still puzzled how it actually came to be. I double-checked it multiple times now with different datasets and it's consistent. Note that the problem is only relevant (in the sense of giving **wrong** values) when combining trigger and stage delay, by themselves it just seems like an odd definition.

@JamesWrigley Do you know whether this was used at MID already (only other really supported instrument)? This will change the values actually, though the impact is *only* the sign if not combining with other delay sources.

### DIFF
--- a/src/extra/components/las.py
+++ b/src/extra/components/las.py
@@ -273,7 +273,7 @@ class OpticalLaserDelay:
     @property
     def _stage_to_time(self) -> float:
         """millimeter to seconds"""
-        return -2e-3 / speed_of_light
+        return 2e-3 / speed_of_light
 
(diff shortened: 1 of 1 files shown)


JamesWrigley (approved):
LGTM! I don't think this component was ever used at MID, we pretty much only use `PumpProbePulses` and `XrayPulses`. The delay is usually recorded through some other Karabo property.

philsmt:
Then I'm relieved, thanks!

PR #231, opened 2024-09-05
Add a Timepix3 component

philsmt:
Timepix3 is a detector used at SQS. While having individual pixels, it is not a frame-based detector, but instead records beginning and the duration of time whenever a pixel is above some defined intensity threshold. Conceptually, it is therefore more similar to a DLD. While it has significantly worse time and spatial resolution than a DLD, it does not suffer from its multi-hit problems - every pixel is independent and it can therefore better cope with lots of signal in very short period of time across the detector.

Raw data is saved as X, Y, ToA (time-of-arrival) and ToT (time-over-threshold) per train. Offline calibration contains a correction step called centroiding, where multiple pixels are combined in space and time to a single detector event called centroid. The resulting data is saved in a similar format, but as structured arrays rather than individual keys.

This component allows to efficiently load both raw and centroid dato into pandas dataframe, similar to the DLD component. The most important step it provides is correlating the detector hits or centroids to EuXFEL pulses via ToA, with the pulse information taken from an EXtra pulse pattern component. Special modes exist to also include detector events occuring outside of pulses.

@bj-s

### DIFF
--- a/src/extra/components/pulses.py
+++ b/src/extra/components/pulses.py
@@ -520,7 +520,8 @@ class PulsePattern:
              DeprecationWarning, stacklevel=2)
         return self.pulse_counts(*args, **kwargs)
 
-    def build_pulse_index(self, pulse_dim='pulseId', include_extra_dims=True):
+    def build_pulse_index(self, pulse_dim='pulseId', include_extra_dims=True,
+                          pulse_dtype=None):
         """Get a multi-level index for pulse-resolved data.
 
         Args:
@@ -529,6 +530,8 @@ class PulsePattern:
             include_extra_dims (bool, optional): Whether to include any
                 additional dimensions of this particular implementation
                 beyond train ID and pulse dimension.
+            pulse_dtype (numpy.dtype, optional): If specified, force
+                casts the pulse dimension to the specified dtype.
 
         Returns:
             (pandas.MultiIndex): Multi-level index covering train ID,
@@ -559,6 +562,9 @@ class PulsePattern:
         if include_extra_dims:
             index_levels.update({name: pulse_ids.index.get_level_values(name)
                                  for name in pulse_ids.index.names[2:]})
+
+        if pulse_dtype is not None:
+            index_levels[pulse_dim] = index_levels[pulse_dim].astype(pulse_dtype)
 
         import pandas as pd
         return pd.MultiIndex.from_arrays(
--- a/src/extra/components/timepix.py
+++ b/src/extra/components/timepix.py
@@ -1,0 +1,618 @@
+
+from collections import defaultdict
+from io import IOBase
+from os import PathLike
+import re
+
+import numpy as np
+import pandas as pd
+
+from extra_data import by_id
+
+
+# Pulse indices used internally for virtual leading and trailing pulses.
+TRAILING_PULSE_INDEX = 2**31 - 1
+LEADING_PULSE_INDEX = -TRAILING_PULSE_INDEX - 1
+
+
+class Timepix3:
+    """Interface to Timepix3 event data.
+
+    Timepix3 is an event-based pixel detectors, generating a stream of
+    events of when and how long a singular pixel is detected. This
+    component gives convenient [pd.dataframe][pandas.DataFrame]-based
+    access to directly acquired raw pixel events, as well as centroided
+    events available after automatic processing. These centroids attempt
+    to group events in time and space that belong to the same particle
+    impact to a single event.
+
+    The component can be initialized with either the raw data source,
+    the virtual calibrated source containing the centroided results or
+    both. Pulse pattern information are necessary in both cases to sort
+    events into pulses, as they are originally recorded by train.
+
+    Args:
+        data (extra_data.DataCollection): Data to access Timepix3 data from.
+        detector (str, optional): Name of the detector, only
+            needed if the data includes more than one. This should
+            be the first part of the source name, i.e. up to the first
+            slash.
+        pulses (extra.components.pulses.PulsePattern, optional): Pulse
+            component to pull pulse information. If omitted, an
+            [XrayPulses][extra.components.XrayPulses] object is
+            constructed from the data.
+    """
+
+    # Only support single-chip detectors for now.
+    _instrument_re = re.compile(
+        r'^(\w{3}_\w+_TIMEPIX)\/(DET|CAL)\/\w+:daqOutput.chip0$')
+
+    def __init__(self, data, detector=None, pulses=None, **kwargs):
+        # Always run detection to potentially find the raw and
+        # centroided data sources depending on passed prefix.
+        self._detector_name, sources = self._find_detector(
+            data, detector or '')
+
+        self._raw_control_src = None
+        self._raw_instrument_src = None
+        self._centroids_control_src = None
+        self._centroids_instrument_src = None
+
+        for source in sources:
+            if '/DET/' in source:
+                self._raw_instrument_src = data[source]
+
+                if (s := source[:source.rfind(':')]) in data.control_sources:
+                    self._raw_control_src = data[s]
+            elif '/CAL/' in source:
+                self._centroids_instrument_src = data[source]
+
+                if (s := source[:source.rfind(':')]) in data.control_sources:
+                    self._centroids_control_src = data[s]
+
+        if self._raw_instrument_src is not None:
+            # For raw data, several keys must be accessed as well as
+            # potentially correlated with centroid data. For efficient
+            # access with pasha, we keep around a minimal DataCollection
+            # selected down to those trains with raw data.
+
+            selection = {self._raw_instrument_src.source: {
+                'data.x', 'data.y', 'data.toa', 'data.tot'}}
+
+            if self._centroids_instrument_src is not None:
+                selection[self._centroids_instrument_src.source] = {
+                    'data.labels'}
+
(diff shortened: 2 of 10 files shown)


src/extra/components/pulses.py
```diff
@@ -520,7 +520,8 @@ def get_pulse_counts(self, *args, **kwargs):
              DeprecationWarning, stacklevel=2)
         return self.pulse_counts(*args, **kwargs)
 
-    def build_pulse_index(self, pulse_dim='pulseId', include_extra_dims=True):
+    def build_pulse_index(self, pulse_dim='pulseId', include_extra_dims=True,
+                          pulse_dtype=None):
```
philsmt, line 524: This feature is added to label the leading and trailing virtual pulses by `-np.inf` and `np.inf`. I tried a couple of different methods, but ended up liking this the most. It's not ideal to have native integers as floats (in the case of `pulseId` or `pulseIndex`), but the alternative would be arbitrary high numbers for the trailing pulse. I tried using `-2` and `-1` respectively, but that became a mess during the alignment process as data is sorted by ToA.

src/extra/components/timepix.py
```diff
@@ -0,0 +1,612 @@
...
+            of either raw or centroided source name.
+        pulses (extra.components.pulses.PulsePattern, optional): Pulse
+            component to pull pulse information. If omitted, an
+            [XrayPulses][extra.components.XrayPulses] object is
+            constructed from the data.
+    """
+
+    # Only support single-chip detectors for now.
+    _instrument_re = re.compile(
+        r'^(\w{3}_\w+_\w+)\/(DET|CAL)\/\w+:daqOutput.chip0$')
```
takluyver, line 45: I assume you've tested this, but the `chip0` comes in the source name, not as a key within the schema? If so, it's a nice surprise if everything works correctly for that in EXtra-data...

Other than that, this source name is very generic. Is there no 'TIMEPIX' or similar part we can expect there?

src/extra/components/timepix.py
```diff
@@ -0,0 +1,612 @@
...
+            timewalk_lut (os.PathLike or None, optional): Path to a file
+                compatible with np.load containing a timewalk LUT to
+                correct ToA, none by default.
+            out_of_pulse_events (bool, optional): Whether to include
+                hits before the first pulse and after the last pulse in
+                virtual pulses labelled -np.inf and np.inf, False by
+                default. If enabled, the pulse index dimension will be
+                forced to float.
+            extended_columns (bool, optional): Whether to include the
+                original time-fo-arrival, readout position and centroid
```
takluyver, line 415:
```suggestion
                original time-of-arrival, readout position and centroid
```

src/extra/components/timepix.py
```diff
@@ -0,0 +1,612 @@
...
+        time-of-arrival is between the beginning of that and the next
+        pulse For the last pulse, the single largest spacing between
+        pulses is chosen while in case of a single pulse in the entire
+        train, all pulses will be assigned to it. The `toa_offset`
+        argument denotes the time-of-arrival at the start of the first
+        pulse of the train.
+
+        Args:
+            pulse_dim ('pulseId' or 'pulseIndex' or 'pulseTime', optional):
+                Label for pulse dimension, pulse ID by default.
```
takluyver, line 403: Wording nitpick: 'label' sounds like it just changes the... label on the pulse index, whereas this actually changes the information in the index. Also, pandas has 'index levels' instead of dimensions. So what about something like...

```suggestion
                Which pulse metadata to use in the index, pulse ID by default.
```

(I also don't think the difference between pulse ID & index is obvious - I had to think about it - but I don't think we want to explain that in every docstring that refers to them.)

takluyver (commented):
Some specific comments, but in general this LGTM

src/extra/components/timepix.py
takluyver: I might put it back, to hedge against getting any other devices with a `daqOutput.chip0` channel. But I don't feel strongly about that, and you probably have a better idea of likely patterns around this than I do.

src/extra/components/timepix.py
```diff
@@ -0,0 +1,612 @@
...
+
+    The component can be initialized with either the raw data source,
+    the virtual calibrated source containing the centroided results or
+    both. Pulse pattern information are necessary in both cases to sort
+    events into pulses, as they are originally recorded by train.
+
+    Args:
+        data (extra_data.DataCollection): Data to access Timepix3 data from.
+        detector (str, optional): Name of the detector, only
+            needed if the data includes more than one. May be any prefix
```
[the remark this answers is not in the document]
philsmt: I think some of them already are implemented by prefix searching, though the intention was to give the entire first part. We could fix this easily by searching for `{prefix}/`.

src/extra/components/timepix.py
```diff
@@ -0,0 +1,612 @@
...
+
+        Returns:
+            centroids (pandas.DataFrame): Centroids.
+        """
+
+        # Prepare aligned data sources.
+        kd = self.centroids_key.drop_empty_trains()
+        tids = kd.train_ids
+        pulses = self._pulses.select_trains(by_id[tids])
+        pids = pulses.pulse_ids()
```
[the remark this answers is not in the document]
philsmt: This would be handled in `_sort_timepix_data` through the `KeyError` branch, though this is indistinguishable from not properly having any pulses. We discussed this potential case of swallowing a warning, but it'd require an additional check for train existance via a call like `PulsePattern.pulse_counts()`. As it is now, data of in such a train would be assigned to the trailing pulse, which is easily identifiable and in the worst case misses the data. There is no risk of falsely assigning data.

src/extra/components/timepix.py
```diff
@@ -0,0 +1,612 @@
...
+
+        # Determine additional pulse indices and values required for
+        # leading and trailing hits, i.e. before and after "pulses", or
+        # create a mask to exclude them.
+        # The indices are used internally and denote those cases with
+        # -1 and INT32_MAX, respectively. The returned index however
+        # is forced to float using -np.inf and +np.inf.
+        # If either virtual pulses are disabled, they are masked out.
+        if out_of_pulse_events:
+            mask = np.ones_like(row_pidx, dtype=bool)
```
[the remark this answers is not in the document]
philsmt: As there's currently no further modification of the mask, I think it's a reasonable optimization. Thanks!

src/extra/components/timepix.py
```diff
@@ -0,0 +1,612 @@
...
+        argument denotes the time-of-arrival at the start of the first
+        pulse of the train.
+
+        Args:
+            pulse_dim ('pulseId' or 'pulseIndex' or 'pulseTime', optional):
+                Label for pulse dimension, pulse ID by default.
+            toa_offset (float, optional): Time-of-arrival offset applied
+                by train in microseconds, 0.0 by default.
+            timewalk_lut (os.PathLike or None, optional): Path to a file
+                compatible with np.load containing a timewalk LUT to
```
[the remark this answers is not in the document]
philsmt: Yes, totally. I didn't think of it much when copying this from the prototype notebooks.

src/extra/components/timepix.py
philsmt: I agree this may not always be obvious, and we should explain this in the `PulsePattern` component. But I'm not entirely sure *pulse metadata* is clearer, how about *pulse annotation*?

src/extra/components/timepix.py
```diff
@@ -0,0 +1,612 @@
...
+            if m is not None and source.startswith(prefix):
+                detectors[m[1]].append(source)
+
+        if len(detectors) > 1:
+            raise ValueError('multiple detectors found, please pass one '
+                             'explicitly:\n' + ', '.join(sorted(detectors)))
+        elif detectors:
+            return next(iter(detectors.items()))
+
+        raise ValueError('no detector found, please pass one explicitly')
```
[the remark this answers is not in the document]
philsmt: That is true.
Coupled with the comment above, I changed the behaviour to exactly expect the Karabo domain or nothing, and raise an appropriate exception.

src/extra/components/timepix.py
philsmt: Coupled with the naming change, it expects the domain now end in `_TIMEPIX`. That at least applies to all known instances.

src/extra/components/timepix.py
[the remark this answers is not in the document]
takluyver: I don't have a particular preference, just that I'd like it to behave the same, as far as possible, for different components.

src/extra/components/timepix.py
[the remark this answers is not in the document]
takluyver: OK, that sounds good. Hopefully missing data from the timeserver is very rare in any case.

src/extra/components/timepix.py
takluyver: Yup, I'm fine with *annotation*. Or *coordinates*, if you want another option. Thinking about it, even *pulse labels* (plural) might be OK.

src/extra/components/timepix.py
takluyver: That sounds good, thanks.

src/extra/components/timepix.py
philsmt: I like coordinates, thanks. That also fits with train ID coordinates in EXtra-data.

takluyver:
Thanks! I think all of my comments before were minor enough that I don't need to look through this again, so if you're happy, go for it.

philsmt:
Thanks for review!

PR #232, opened 2024-09-06
Move XGM mock devices to their own module in .mockdata

philsmt:
@JamesWrigley If you don't mind?


JamesWrigley (approved):
Not at all, LGTM!

PR #236, opened 2024-09-13
Fix initializing PumpProbePulses with an instrument string

philsmt:
This was broken at some point during or after development when static properties got moved to helper functions, and got missed in the tests. This PR adds a test for it.

I also merged helper functions in `extra.components._common` into `extra.components.utils`... I suspect we added such modules in different PRs 🤷


src/extra/components/utils.py
```diff
@@ -7,6 +7,33 @@
     3: {'SA3', 'LA3', 'SCS', 'SQS', 'SXP'}
 }
 
+
+def _identify_instrument(data, threshold=0.5):
```
takluyver, line 11: Looks like this function is completely uncovered by tests. It would be nice to exercise the most-common-prefix pathway, directly or indirectly. It doesn't need to be in this PR, though.

src/extra/components/_common.py
```diff
@@ -34,10 +34,6 @@ def _instrument_to_sase(instrument):
         return 3
 
 
-def _identify_sase(data):
-    return _instrument_to_sase(_identify_instrument(data))
-
-
 def _select_subcomponent_trains(src, keys, dst=None):
```
takluyver, line 37: Did you mean to get rid of the rest of this module & move it into `.utils`?

philsmt: Not quite, I meant to merge anything that is **used** into `.utils`. That particular function `extra.components._common._identify_sase` was actually never used, but there's an `extra.components.utils.identify_sase` that is. It is meant to keep the latter and drop the former.

src/extra/components/utils.py
philsmt: Done in 7406154abad848374b7aecfacaf2a6298573c2b5

src/extra/components/_common.py
takluyver: That makes sense, but I'm wondering why you haven't removed this file entirely?

philsmt: Because I didn't `git add` the removal...

takluyver: Aha, thank-you!

takluyver:
LGTM

philsmt:
Thanks!

PR #242, opened 2024-09-24
Improve extrapolation exceptions in PumpProbePulses 

philsmt:
`PumpProbePulses` can anchor PPL pulses to FEL pulses via three different modes: `bunch_table_position`, `bunch_table_offset` and `pulse_offset`. While the first one always works, the second requires at least one while the third requires at least two pulses. Thus, an exception is raised if there are insufficient pulses in a given train to satisfy this mode. With `extrapolate=True` (default), an attempt is made to instead use the last working pattern.

This PR clarifies the exceptions, as this caused confusion in the field. It clearly distinguishes now between missing _one_ or _two_ pulses, and explains why. In addition it makes clear that it tried but failed to extrapolate, e.g. if it never saw a usable train yet.

In addition, it adds a special case for `pulse_offset=0` to also work with a single pulse by replacing it internally with `bunch_table_offset=0`.

@bermudei

### DIFF
--- a/src/extra/components/pulses.py
+++ b/src/extra/components/pulses.py
@@ -1288,6 +1288,8 @@ class PumpProbePulses(XrayPulses, OpticalLaserPulses):
         self._pulse_offset = None  # Allowed to be float!
         self._extrapolate = extrapolate
 
+        # Logic should always check these modes in exactly this order,
+        # see below.
         if bunch_table_position is not None:
             self._bunch_table_position = int(bunch_table_position)
         elif bunch_table_offset is not None:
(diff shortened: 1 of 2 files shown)


philsmt:
codecov seems to complain about `__repr__`, I somehow hesitate to do string grepping there.

src/extra/components/pulses.py
```diff
@@ -1298,6 +1300,14 @@ def __init__(self, data, source=None, instrument=None, *,
             raise ValueError('must specify one of bunch_table_position, '
                              'bunch_table_offset, pulse_offset')
 
+        if self._pulse_offset == 0:
+            # Implement this case via the bunch_table_offset mechanism
+            # to enable it to work even if there is only a single FEL
+            # pulse in any given train.
+            # Implemenmtation-wise the modes are always checked in the
```
JamesWrigley, line 1307: Nitpicking:
```suggestion
            # Implementation-wise the modes are always checked in the
```

JamesWrigley (changes requested):
Nice :+1: This always trips me up.

src/extra/components/pulses.py
philsmt: Oops, thanks. Also found another typo while fixing it.

JamesWrigley (approved):
LGTM!

philsmt:
Thanks!

PR #245, opened 2024-09-26
Fill pulse traces with NaN/-1 if data is too short

philsmt:
When using the pulse separation feature in the `AdqRawChannel` component, currently an exception is thrown if the underlying train trace is too short for all pulses. This is not very useful however, as there's little way for users to fix that in any way, e.g. only get data for those pulses that fit, or reduce the pattern asked for.

Rather than throwing an exception, this PR instead fills the missing samples with `NaN` (for floating types, most likely) or `-1` (in case of integer types, only with non-default settings). This maintains the information which data is missing, and the returned array can easily be truncated if desired.

### DIFF
--- a/src/extra/components/_adq.pyx
+++ b/src/extra/components/_adq.pyx
@@ -4,6 +4,7 @@
 from cython cimport numeric
 from libc.limits cimport INT_MAX
 from libc.string cimport memcpy
+from numpy.math cimport NAN
 
 
 def _reshape_flat_pulses(
@@ -15,11 +16,15 @@ def _reshape_flat_pulses(
     if pulse_ids.shape[0] > out.shape[0]:
         raise ValueError('pulse output buffer smaller than pulse count')
 
-    cdef int start, end, i, j = -1, \
+    cdef int start, samples_to_copy, i, j = -1, k, \
         cur_pid, prev_pid = INT_MAX, pid_offset = 0, \
         num_trains = traces.shape[0], trace_len = traces.shape[1]
 
-    cdef size_t bytes_per_pulse = samples_per_pulse * sizeof(traces[0, 0])
+    # Prepare the placeholder value for float and int types, using the
+    # fact that NAN != NAN for the true float value while becoming equal
+    # after integer conversion.
+    cdef numeric placeholder_val = -1 if <numeric>NAN == <numeric>NAN \
+        else <numeric>NAN
 
     for i in range(pulse_ids.shape[0]):
         cur_pid = pulse_ids[i]
@@ -34,9 +39,15 @@ def _reshape_flat_pulses(
 
         prev_pid = cur_pid
         start = first_pulse_offset + (cur_pid - pid_offset) * clock_ratio
-        end = start + samples_per_pulse
 
-        if end > trace_len:
-            raise ValueError(f'trace too short for pulse at {start}:{end}')
+        # The end of this pulse is beyond the trace length, try to
+        # copy what remains and fill with placeholder as needed.
+        samples_to_copy = min(max(trace_len - start, 0), samples_per_pulse)
 
-        memcpy(&out[i, 0], &traces[j, start], bytes_per_pulse)
+        # Copy if data available.
+        memcpy(&out[i, 0], &traces[j, start],
+               samples_to_copy * sizeof(traces[0, 0]))
+
+        # Fill in the rest of this pulse with the placeholder value.
+        for k in range(samples_to_copy, samples_per_pulse):
+            out[i, k] = placeholder_val
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -34,6 +34,7 @@ Changed:
   use by default (!221).
 - [`fit_gaussian()`][extra.utils.fit_gaussian] has a new `A_sign` parameter to
   specify the expected orientation of the peak (!222).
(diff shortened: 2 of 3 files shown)


src/extra/components/_adq.pyx
```diff
@@ -37,6 +44,16 @@ def _reshape_flat_pulses(
...
+            if start < trace_len:
+                # There is still some data to copy into this pulse.
+                memcpy(&out[i, 0], &traces[j, start],
+                       (trace_len - start) * sizeof(traces[0, 0]))
+
+            # Fill in the rest of this pulse with the placeholder value.
+            for k in range(max(trace_len - start, 0), samples_per_pulse):
+                out[i, k] = placeholder_val
+        else:
+            memcpy(&out[i, 0], &traces[j, start], bytes_per_pulse)
```
egorsobolev, line 59: Looks it possible to make a bit simple:

```suggestion
        # The end of this pulse is beyond the trace length, try to
        # copy what remains and fill with placeholder as needed.
        samples_to_copy = min(max(trace_len - start, 0), samples_per_pulse)

        # copy if data available
        memcpy(&out[i, 0], &traces[j, start], samples_to_copy * sizeof(traces[0, 0]))

        # Fill in the rest of this pulse with the placeholder value.
        for k in range(samples_to_copy, samples_per_pulse):
            out[i, k] = placeholder_val
```

src/extra/components/_adq.pyx
```diff
@@ -37,6 +44,16 @@ def _reshape_flat_pulses(
         end = start + samples_per_pulse
 
         if end > trace_len:
```
egorsobolev, line 46:
```suggestion
```

philsmt:
Anything else @egorsobolev ?

src/extra/components/_adq.pyx
```diff
@@ -34,9 +41,15 @@ def _reshape_flat_pulses(
...
+        samples_to_copy = min(max(trace_len - start, 0), samples_per_pulse)
+
+        # Copy if data available.
+        memcpy(&out[i, 0], &traces[j, start],
+               samples_to_copy * sizeof(traces[0, 0]))
 
-        memcpy(&out[i, 0], &traces[j, start], bytes_per_pulse)
+        # Fill in the rest of this pulse with the placeholder value.
+        for k in range(samples_to_copy, samples_per_pulse):
+            out[i, k] = placeholder_val
```
egorsobolev, line 55: `memset` is fill memory with `0xff`, which is decoded as `-1` or `nan` depending on type. You may simplify `nan` placeholder value selection and padding:
```suggestion
        memset(&out[i, samples_to_copy], -1, (samples_per_pulse - samples_to_copy) * sizeof(out[0, 0]))
```

src/extra/components/_adq.pyx
```diff
@@ -15,12 +16,18 @@ def _reshape_flat_pulses(
     if pulse_ids.shape[0] > out.shape[0]:
         raise ValueError('pulse output buffer smaller than pulse count')
 
-    cdef int start, end, i, j = -1, \
+    cdef int start, samples_to_copy, i, j = -1, k, \
         cur_pid, prev_pid = INT_MAX, pid_offset = 0, \
         num_trains = traces.shape[0], trace_len = traces.shape[1]
 
     cdef size_t bytes_per_pulse = samples_per_pulse * sizeof(traces[0, 0])
```
egorsobolev, line 24: Looks, it is not used anymore
```suggestion
```

src/extra/components/_adq.pyx
```diff
@@ -15,12 +16,18 @@ def _reshape_flat_pulses(
...
 
+    # Prepare the placeholder value for float and int types, using the
+    # fact that NAN != NAN for the true float value while becoming equal
+    # after integer conversion.
+    cdef numeric placeholder_val = -1 if <numeric>NAN == <numeric>NAN \
+        else <numeric>NAN
+
```
egorsobolev, line 30:
```suggestion
```
Not needed if you use `memset`

src/extra/components/_adq.pyx
```diff
@@ -4,6 +4,7 @@
 from cython cimport numeric
 from libc.limits cimport INT_MAX
 from libc.string cimport memcpy
+from numpy.math cimport NAN
```
egorsobolev, line 7: If you decide to go with `memset`
```suggestion
from libc.string cimport memset
```

src/extra/components/_adq.pyx
egorsobolev: `k` as well in this case

takluyver:
LGTM, give or take Egor's memset suggestion; I don't have a preference on that.

egorsobolev:
LGTM, except these suggestions. For me, the `memset` implementation looks typical in c. But here it is up to you.

src/extra/components/_adq.pyx
philsmt: Some googling suggests that while this will work most of the time, it's in the realm of undefined behaviour. So I'd rather not.

philsmt:
Thanks for review! Using `memset` to initialize `float` or `double` arrays is considered undefined behaviour (technically even setting them to zero), so I'd rather not. No doubt it's a bit faster, but unlikely to make a huge difference.

egorsobolev:
> Using `memset` to initialize `float` or `double` arrays is considered undefined behaviour (technically even setting them to zero), so I'd rather not. 

If there any system nowadays which uses a non-IEEE 754 floating point format? As far as I know the last representatives died out 20 years ago.

philsmt:
It really shouldn't, but google *float all bits 0* if you want to join the fun.

PR #250, opened 2024-10-16
Catch reported number of Timepix events exceeding buffer size 

philsmt:
Raw Timepix3 data is saved as ragged arrays with a fixed (but configurable) buffer size with the actual number of events recorded in a `data.size` key. While the device should prevent this from ever exceeding the buffer size, there is some data floating around where this happened. This PR makes sure the component can handle that, and in addition prints a warning as this is unlikely to be desired from data quality point of view.

Also includes a minor change to move the `centroid_size` column from the set of extended columns to default columns. Apart from someone iterating over columns (please don't), this should not break any existing columns.

@bj-s

### DIFF
--- a/src/extra/components/timepix.py
+++ b/src/extra/components/timepix.py
@@ -551,8 +551,9 @@ class Timepix3:
                 False by default. If enabled, the pulse index dimension
                 will be forced to float.
             extended_columns (bool, optional): Whether to include the
-                original average and maximal time-over-threshold,
-                size and label for each centroid, False by default.
+                average and maximal time-over-threshold as well
+                as original time-of-arrival and labels for each
+                centroid, False by default.
             parallel (int or None, optional): Nunmber of parallel
                 processes to use, by default 10 or a quarter of all cores
                 whichever is lower. Any non-positive value or 1 disable
@@ -621,14 +622,17 @@ class Timepix3:
         # Build the data frame with centroids and the prepared index.
         df = pd.DataFrame.from_records(
             centroids.ravel()[centroids_rows][mask], index=index,
-            exclude=(['size', 'tot_avg', 'tot_max', 'toa']
+            exclude=(['tot_avg', 'tot_max', 'toa']
                      if not extended_columns else []))
         df.insert(2, 't', centroids_tof[mask])
+        df.rename(columns={'size': 'centroid_size'}, inplace=True)
 
         if extended_columns:
-            df.rename(columns={'size': 'centroid_size'}, inplace=True)
+            # Re-order centroid size.
+            df.insert(4, 'centroid_size', df.pop('centroid_size'))
+
             df['toa'] = centroids_toa[mask]  # Overwrite with modified data.
-            df.insert(6, 'toa', df.pop('toa'))  # Re-order ToA
+            df.insert(7, 'toa', df.pop('toa'))  # Re-order ToA
             df.insert(3, 'tot', df.pop('tot'))  # Re-order ToT.
             df['label'] = centroids_label[mask]
 
--- a/src/extra/components/timepix.py
+++ b/src/extra/components/timepix.py
@@ -441,6 +441,21 @@ class Timepix3:
         hits_per_train = self.raw_size_key.ndarray()
         num_hits = int(hits_per_train.sum())
 
+        if (hits_per_train > self.raw_x_key.entry_shape[0]).any():
+            # data.size may actually record more pixel events than can
+            # be placed into array buffer. The device is supposed to
+            # prevent this, but there is some commissioning data
+            # affected by it.
+
+            diff = hits_per_train - self.raw_x_key.entry_shape[0]
+            max_deviation = diff.max()
+            num_deviations = (diff > 0).sum()
+
+            from warnings import warn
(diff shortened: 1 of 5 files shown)


src/extra/components/timepix.py
```diff
@@ -441,6 +441,21 @@ def pixel_events(self, pulse_dim='pulseId', toa_offset=0.0,
         hits_per_train = self.raw_size_key.ndarray()
         num_hits = int(hits_per_train.sum())
 
+        if (hits_per_train > self.raw_x_key.entry_shape[0]).any():
+            # data.size may actually record more pixel events than can
+            # be placed into array buffer. The device is supposed to
+            # prevent this, but there is some commissioning data
```
takluyver, line 447: Does the fixed device still record somewhere that some hits were dropped because of the buffer size? It seems worth having some way to surface that.

tests/test_components_timepix.py
```diff
@@ -162,6 +167,23 @@ def test_timepix3_pixel_events(mock_sqs_timepix_run):
...
+        res = orig_ndarray(self, *args, **kwargs)
+
+        if self.source == tpx.raw_instrument_src.source and self.key == 'data.size':
+            res[res.argmax()] += tpx.raw_x_key.entry_shape[0]
+
+        return res
+
+    from extra_data import KeyData
+    orig_ndarray = KeyData.ndarray
+    monkeypatch.setattr(KeyData, 'ndarray', _ndarray)
```
takluyver, line 182: This feels like we're doing something pretty complex to test something very simple. I don't see any specific problems it's likely to cause, so feel free to carry on, but... I would probably do this with the `mockdata` framework. Maybe we need some more machinery to make it easier to modify a copy of a file for tests? :thinking:

takluyver:
LGTM

src/extra/components/timepix.py
philsmt: Yes, as I understand from @bj-s the device is supposed to go into error in this situation.

tests/test_components_timepix.py
philsmt: Yes, I'm not particularly proud of this hack, but in the end it seemed to be the most stable to *just* change the returned values. I started with `mockdata`, but subtle things were breaking on other operations and index alignment. In the end EXtra components like this heavily rely on sane data and metadata.

It would be neat to be able to override values for a single test. The EXtra-data test framework is great, but that case is tricky to do. I did not want to add another device with bogus data just for the sake of this test.

takluyver: We've got some tests which do this already, with fixtures which make a fresh copy of a regular example file and then modify it:

https://github.com/European-XFEL/EXtra-data/blob/479ea37389d43515a41736aaa9a62d505cd0ed67/extra_data/tests/test_bad_trains.py#L15-L52

Would something like that work? It's a bit clumsy because you have to explicitly put the full dataset path there to modify it, but that's not exactly difficult.

We could simplify this by making the fixtures for test data function-scoped, so each test gets its own copy of the files anyway and can modify them in place. I made them session scoped to start with for performance, but creating a few temp files shouldn't be that slow. :shrug:

philsmt: No, I think that'll work nicely. Thanks for the hint, TIL.

philsmt: What do you think of 03f19ad808c1a1fd7308548b585328f3e58f6ace? I decided to make a new function-scoped fixture around the original module-scoped data that changes the value back and forth. This should avoid having an entire second run in tmp.

takluyver: I'd rather copy and modify than have a fixture that changes the value back afterwards, to tell you the truth - that sounds like premature optimisation. Copying a small file on a tmpfs should be pretty fast.

If we wanted to do this with directories with many files, we could get smart about linking the ones that we don't want to change.

tests/conftest.py
```diff
@@ -127,3 +127,21 @@ def mock_sqs_timepix_directory():
...
 def mock_sqs_timepix_run(mock_sqs_timepix_directory):
     yield RunDirectory(mock_sqs_timepix_directory)
+
+
+@pytest.fixture(scope='function')
+def mock_timepix_exceeded_buffer_run(mock_sqs_timepix_directory):
+    run = RunDirectory(mock_sqs_timepix_directory).deselect('SQS_EXTRA*')
+    tpx_root = 'INSTRUMENT/SQS_EXP_TIMEPIX/DET/TIMEPIX3:daqOutput.chip0'
+
+    with h5py.File(run.files[0].filename, 'r+') as f:
```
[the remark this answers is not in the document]
philsmt: Just noticed, it didn't happen before when I ran the test by itself. Already fixed in 0a08af96cfb7a2fc49b64fbf787ed1375e58ba5f

tests/test_components_timepix.py
philsmt: Hu, copying a file back-and-forth is indeed more clever, good idea.

philsmt: Like c48ac4d5cf9681416fcd560857700a88336e8fb0?

takluyver: Yup, LGTM, thanks!

PR #259, opened 2024-11-20
Fix pulse slicing code to advance to next train on same pulse ID

philsmt:
The native pulse slicing code compares subsequent pulse IDs to advance from train to train. In the case of a single pulse, this value never changes, thus reading the same (first) train over and over.


takluyver:
LGTM; could have a test, but I don't think it's essential

philsmt:
I fully agree, but the [testing infrastructure](https://github.com/European-XFEL/EXtra/pull/215) for this component requries some minor polishing. This bug is too serious to have it wait, though. Thanks!

PR #260, opened 2024-11-20
Relax contourpy dependecy to avoid version conflicts

philsmt:
https://github.com/European-XFEL/EXtra/pull/256 appears to have introduced a dependency problem, as this version requires Python 3.10 and above.


takluyver:
LGTM

PR #262, opened 2024-11-23
Allow passing a figure size to Scan.plot()

JamesWrigley:


philsmt:
Why single out `figsize` and not use the already existing mechanism to pass an `Axes` object?

JamesWrigley:
(if there's no objections I'll merge this at the end of the week)

takluyver:
Fine by me :+1:

PR #263, opened 2024-11-30
Fix alignment between DLD data and external pulse information

philsmt:
While most people use the DLD's internal pulse information, the component also supports supplying a `PulsePattern` object from the outside. Then, it may be necessary to align them.

The code supposed to handle that was unfortunately not working properly, as restricting `entry_counts` to only those rows with data before can easily cause the condition below to not properly activate. As a matter of fact, the test written for exactly this line hapened to pass because while it contained more trains, those had no pulses...

@bj-s


bj-s:
LGTM

PR #269, opened 2024-12-09
Only average over trains dimension in Scan.bin_by_steps()

takluyver:
This came up at FXE: if you have an array with multiple values per train, such as a spectrum, you naturally want an average spectrum per scan step, but `scan.bin_by_steps(arr)` reduces it down to a single number per scan step instead.

This could theoretically break code that relies on it collapsing extra dimensions, but I'm not sure when you'd want that, and it's very easy to do further averaging on the results from this if you need to. So I hope we can just consider it a bug and fix it. :crossed_fingers: 

cc @bermudei

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -27,6 +27,9 @@ Fixed:
 - [`Scan.plot_bin_by_steps()`][extra.components.Scan.plot_bin_by_steps] would
   previously ignore the `title`/`xlabel`/`ylabel` arguments, now it actually
   uses them (!237).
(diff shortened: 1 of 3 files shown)


JamesWrigley (changes requested):
Won't this also remove the extra dimension names and such? It'd be nice to preserve those if possible. This should also have a test.

JamesWrigley (approved):
LGTM!

takluyver:
Thanks!

PR #275, opened 2024-12-17
Fix too restrictive auto detection in Timepix component

### LINKED ISSUE #252: Timepix3 component does not work with old data source naming
The `Timepix3` component does not accept the data sources used with the initial (now old) implementation of the Timepix3 camera.

1) the regular expression used to identify the data source is not accounting for those names
2) the class requires `/DET/` to be part of the source name to identify it as pixelated data

data sources are:
* 'SQS_AQS_CAM/CAM/TIMEPIX3' (control)
* 'SQS_AQS_CAM/CAM/TIMEPIX3:daqEventOutput' (pixelated data)

see e.g. p2873, run 118.

### LINKED ISSUE #273: Timepix3 component fails with old calibration/centroiding result
The Timepix3 component expects the property `data.labels` to be present in calibrated data, when loading pixelated data with extended colums. However this is not necessarily the case for older calibration/centroiding results.

When checking whether calibration results are present, it should be further checked whether the key `data.labels` is present in the data.

e.g. there:
```
if extended_columns and self._centroids_instrument_src is not None:
            hits_label = psh.alloc(shape=num_hits, dtype=np.int32)
        else:
            hits_label = None
```
see e.g. p3408 r50

philsmt:
The auto detection of sources in the Timepix component was initially fairly restrictive to transparently support both raw and centroided data at the same time, but this is causing some [issues in the field](https://github.com/European-XFEL/EXtra/issues/252).

This MR allows to pass a tuple of explicit (instrument!) source names for the raw and centroided data, respectively. As before, it will attempt to add the correspondig control data automatically. In addition, it relaxes the restriction on expected source names to also work with `/CAM/`.

While at it, I also fixed [another bug](https://github.com/European-XFEL/EXtra/issues/273) related to older centroiding data not having the `data.labels` key, which causes an unchecked exception if tried.

@bj-s

Closes #252
Closes #273

### DIFF
--- a/src/extra/components/timepix.py
+++ b/src/extra/components/timepix.py
@@ -46,7 +46,7 @@ class Timepix3:
 
     # Only support single-chip detectors for now.
     _instrument_re = re.compile(
-        r'^(\w{3}_\w+_TIMEPIX)\/(DET|CAL)\/\w+:daqOutput.chip0$')
+        r'^(\w{3}_\w+_TIMEPIX)\/(CAM|DET|CAL)\/\w+:daqOutput.chip0$')
 
     def __init__(self, data, detector=None, pulses=None, **kwargs):
         # Always run detection to potentially find the raw and
@@ -60,7 +60,7 @@ class Timepix3:
         self._centroids_instrument_src = None
 
         for source in sources:
-            if '/DET/' in source:
+            if '/DET/' in source or '/CAM/' in source:
                 self._raw_instrument_src = data[source]
 
                 if (s := source[:source.rfind(':')]) in data.control_sources:
--- a/src/extra/components/timepix.py
+++ b/src/extra/components/timepix.py
@@ -80,7 +80,7 @@ class Timepix3:
             selection = {self._raw_instrument_src.source: {
                 'data.x', 'data.y', 'data.toa', 'data.tot'}}
 
-            if self._centroids_instrument_src is not None:
+            if self._has_centroid_labels():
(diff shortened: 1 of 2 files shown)


bj-s:
Allowing also .../CAM/... with the regular expression is not enough to detect the old source naming, also instead of `:daqOutput.chip0` it would would be `:daqEventOutput` for raw data or `:output` for centroided/CAL data.

philsmt:
Yes, for the legacy data I would indeed expect to use the explicit source names. I added the domain there mostly in case someone might use it again with the newer device.

src/extra/components/timepix.py
```diff
@@ -124,26 +130,65 @@ def _prepare_pasha(parallel):
...
+
+            if prefix_or_source:
+                raise ValueError(f'no detector sources found for '
+                                 f'{prefix_or_source}, please pass explicit '
+                                 f'source name(s)')
+            else:
+                raise ValueError('no detector detected, please narrow the '
+                                 'search with the `detector` argument')
+
+        raise ValueError('detector may be a string, tuple of two strings or '
```
takluyver, line 190:
```suggestion
        raise TypeError('detector may be a string, tuple of two strings or '
```

takluyver:
Other than that, LGTM

src/extra/components/timepix.py
philsmt: Fair point, I tend to use `ValueError` too much.

philsmt:
Thanks for review!

PR #284, opened 2025-01-13
Added analysis code used to calibrate Cookiebox and grating spectrometer

daniloefl:
This PR includes the addition of an `ingredients` directory in EXtra, in which self-contained analysis steps are to be performed. Two elements have been added:
1. `CookieboxCalibration` calibrates eTOF data from a calibration run (or a union thereof).
2. `Grating2DCalibration` calibrates grating spectrometer data from a background run and a calibration run (or a union thereof).

## `Grating2DCalibration`

This device takes a background run, and monochromated runs, subtracts the background average data, sorts the monochromated runs by the mono energy and optionally rotates them.

After this step, it fits a line mapping the sample index of the grating data to energy.

The `apply` function applies the background subtraction and rotation for a new run and returns an `xarray.Dataset` containing the transformed data with an energy axis.

## `CookieboxCalibration`

eTOFs provide a 1D trace for all pulses, which can be transformed into traces per pulse by the `AdqRawChannel` Extra component. However, the trace sample axis is meaningless and needs to be converted into energy using a non-linear function. Usually one data analysis run is taken in which the undulator energy is scanned to obtain this map, which is then applied in the actual analysis run.

The objective is to take a calibration run (where an energy scan is made) and from that run calculate a) the map to be used to interpolate eTOF energy axis from "sample number" into eV; and b) calculate the transmission correction, which corrects the appropriate intensity of the data, given that the eTOF does not measure electrons with equal probability in all energy ranges.

The concrete steps taken when the object is initialized (`obj = CookieboxCalib(calib_run)`) from a calibration run are:
- From a calibration run, derive a sample number (proportional to time-of-flight) to energy non-linear map.
- Estimate the transformation Jacobian and use it to correct the spectrum intensity distribution, so that the probability  of photo-electrons observed in a range of energies agrees with the probability in the corresponding time-of-flight range.
- Calculate a transmission correction due to the eTOF quantum efficiency as "photo-electron integral/Auger+Valence integral".
- Calculate a normalization correction due to the pulse energy as taken from the XGM: "Auger+Valence integral/XGM pulse energy".

The concrete steps taken when the `obj.apply(other_run)` is called with a run to be calibrated are:
- Use derived non-linear map to interpolate energy axis per eTOF
- Subtract per energy offset per eTOF.
- Scale data by the inverse of the absolute value of the Jacobian per eTOF, following the change-of-variable theorem in statistics.
- Divide by the transmission per eTOF.
- Divide by the normalization correction.

The variables calculated can be visualized using `obj.plot_diagnostics()`  and similar other functions for validation and other cross-checks.

The current code works with fine-tuning of the constructor arguments `first_pulse_offset`, `auger_start_roi`, `start_roi` and `stop_roi`. There is code already to make it detect them automatically if `None` is set, but it cannot be guaranteed to always work.

Further steps proposed previously by @philsmt which have not been applied yet:
- Split the constructor to allow some of the steps to be repeated. This has been done by creating a series of `update_...` member functions and appropriate setters and getters. The setters and getters update a variable and call the update functions to redo calculations after resetting the corresponding variables.
- Accept AdqRawChannel as input. This is possible by setting them as the value of the map given in `tof_settings`. One could then set `tof_settings={'1_A': AdqRawChannel(...), '1_C': AdqRawChannel(...)}` to us this functionality. Providing a run is still necessary, as other objects (XGM, undulator reading and scans) are needed.
- Possibly use pasha if it shows a benefit when rewriting the code. This was not done, but only because it requires previous knowledge of the size of the structures. I would prefer to keep it as it is, instead of adding complexity on the discovery of it for now. I may change my mind later, but I think this is not the main performance hog at the moment.

### DIFF
--- a/offline_tests/integrated_test_recipes_grating.py
+++ b/offline_tests/integrated_test_recipes_grating.py
@@ -16,9 +16,9 @@ def test_grating_2d_calibration():
 
     calib = Grating2DCalibration(angle=-10.5)
     # do calibration
-    calib.setup(bkg_run["SQS_DIAG3_BIU/CAM/CAM_6:daqOutput", "data.image.pixels"],
-                mono_run["SQS_DIAG3_BIU/CAM/CAM_6:daqOutput", "data.image.pixels"],
-                Scan(mono_run["SA3_XTD10_MONO/MDL/PHOTON_ENERGY", "actualEnergy"], resolution=1e-3))
+    calib.setup(mono_run["SQS_DIAG3_BIU/CAM/CAM_6:daqOutput", "data.image.pixels"],
+                bkg_run["SQS_DIAG3_BIU/CAM/CAM_6:daqOutput", "data.image.pixels"],
+                Scan(mono_run["SA3_XTD10_MONO/MDL/PHOTON_ENERGY", "actualEnergy"], resolution=1))
 
     calib.to_file('grating_calib.h5')
     cal_read = Grating2DCalibration.from_file('grating_calib.h5')


daniloefl:
PS: A test could be written by comparing how the calibration applied to a run differs from the results of the code applied previously to the same run. The issue is that we don't have access to data in those tests.
Other tests are possible, but without data, it is not clear if they help much.

daniloefl:
The `CookieboxCalib` and `GratingCalib` ingredients can be tested using the notebooks in:
- `/gpfs/exfel/data/scratch/danilo/EXtra/Test CookieboxCalib.ipynb`
- `/gpfs/exfel/data/scratch/danilo/EXtra/Test GratingCalib.ipynb`

src/extra/recipes/base.py
```diff
@@ -0,0 +1,75 @@
+
+import h5py
+from dataclasses import dataclass, asdict, is_dataclass
+
+def save_dict(h5grp, obj: Dict[str, Any]):
```
JamesWrigley, line 6: What do you think about saving the settings in an xarray Dataset? That allows for storing quite rich arrays with coordinates/attributes, and we wouldn't have to write any code to save/load it from an HDF5 file.

src/extra/recipes/cookiebox.py
```diff
@@ -0,0 +1,1047 @@
...
+
+def search_offset(trace: np.ndarray, sigma: float=20):
+    """
+    Find highest peaks in the 1D trace.
+    """
+    axis = np.arange(len(trace))
+    gaussian = np.exp(-0.5*(axis - np.mean(axis))**2/sigma**2)
+    gaussian /= np.sum(gaussian, axis=0, keepdims=True)
+    # apply it to the data
+    smoothened = fftconvolve(trace, gaussian, mode="same", axes=0)
```
JamesWrigley, line 49: Is this doing the same thing as https://docs.scipy.org/doc/scipy/reference/generated/scipy.ndimage.gaussian_filter1d.html?

src/extra/recipes/cookiebox.py
```diff
@@ -0,0 +1,1047 @@
...
+    peak_idx = np.argmax(smoothened)
+    return peak_idx - 200
```
JamesWrigley, line 51: Where is 200 coming from?

src/extra/recipes/cookiebox.py
```diff
@@ -0,0 +1,1047 @@
...
+    return p
+
+def model(ts,c,e0,t0):
+    """
+    Model to fit energy as a function of the sample axis.
+    """
+    # e = e0 + c/(ts - t0)**2
+    # ts = np.sqrt(c/(e - e0)) + t0
+    # dt/de = d/de [ sqrt(c/e) ] = d/du sqrt(u) d/de [c/e] = -0.5*c/sqrt(c/e) * 1/(e-e0)^2 
+    # dt/de = -0.5*c/(np.sqrt(c/(e - e0)))/(e - e0)**2
```
JamesWrigley, line 67: It's not quite clear to me if this code is for documentation or a previous implementation that can be removed?

src/extra/recipes/cookiebox.py
```diff
@@ -0,0 +1,1047 @@
...
+    """
+    Fit model using linear regression for fixed t0.
+    """
+    res = linregress(1/np.asarray(ts-t0)**2,np.asarray(energies))
+    c = res.slope
+    e0 = res.intercept
+    return c,e0,t0
+
+def norm_diff_t0(ts,energies,t0):
+    #print(ts,'fi',energies)
```
JamesWrigley, line 84: Nitpicking, I think this can be removed (same for the other print statements).

src/extra/recipes/cookiebox.py
```diff
@@ -0,0 +1,1047 @@
...
+    else:
+        raise Exception('fit did not converge.')
+
+
+def calc_mean(itr: Tuple[int, int], scan: Scan, xgm_data: xr.DataArray, tof: Dict[int, AdqRawChannel], xgm_threshold: float, filter_length: int) -> xr.DataArray:
+    """A function doing the actual calibration, per tof, per energy index."""
+    tof_id, energy_id = itr
+    energy, train_ids = scan.steps[energy_id]
+    #xgm = XGM(run.select_trains(by_id[list(train_ids)]), "SA3_XTD10_XGM/XGM/DOOCS")
+    #sel_xgm_data = xgm.pulse_energy().stack(pulse=('trainId', 'pulseIndex'))
```
JamesWrigley, line 110: Nitpicking, I think this can be deleted.

src/extra/recipes/cookiebox.py
```diff
@@ -0,0 +1,1047 @@
...
+        self.update_roi()
+        self.update_fit_result()
+        self.update_calibration()
+
+    @property
+    def xgm_source(self) -> str:
+        return self._xgm_source
+
+    @xgm_source.setter
+    def xgm_source(self, value: str):
```
JamesWrigley, line 386: There's a lot of these getters and setters :sweat_smile: I'm not sure how this class is typically used, but would it make sense to require that some/all parameters are passed in the constructor, and changing any of them requires creating a new object? In that sense the class would become a bit more 'read-only', which I think is simpler than allowing mutation everywhere.

src/extra/recipes/cookiebox.py
```diff
@@ -0,0 +1,1051 @@
...
+                     for tof_id, (digitizer, channel) in self._tof_settings.items()}
+        all_trace = xr.concat([v.select_trains(np.s_[:n_trains]).train_data(roi=np.s_[:n_samples]) for v in tof.values()],
+                              pd.Index(list(tof.keys()), name="tof"))
+        data = -all_trace.mean(["tof", "trainId"])
+        peaks = search_offset(data)
+        # smooth = scipy.ndimage.gaussian_filter1d(data, sigma=50)
+        # thr = np.median(smooth)
+        # peaks, _ = scipy.signal.find_peaks(smooth, height=thr)
+        # if len(peaks) == 0:
+        #     raise RuntimeError(f"Unable to find peaks in averaged data.")
```
JamesWrigley, line 660: Nitpicking, can this be deleted?

src/extra/recipes/grating.py
```diff
@@ -0,0 +1,172 @@
...
+    """
+    Calibrate a 2D grating spectrometer.
+
+    Args:
+      grating_source: Where to read the grating data from.
+      angle: The rotation angle in degrees if the camera is not aligned.
+      energy_source: Where to read the undulator energy from.
+    """
+    def __init__(self,
+                 grating_source: str='SQS_DIAG3_BIU/CAM/CAM_6:daqOutput',
```
JamesWrigley, line 34: Is there only one of these at XFEL?

src/extra/recipes/cookiebox.py
```diff
@@ -0,0 +1,1051 @@
...
+                    continue
+            plt.plot(self.energy_axis, jacobian, lw=2, label=f"eTOF {tof_id}")
+        plt.xlabel("Energy [eV]")
+        plt.ylabel("Jacobian [1/eV]")
+        plt.legend()
+        plt.show()
+
+    def plot_all_normalizations(self, tof_ids: Optional[List[int]]=None):
+        """
+        Plot all normalizations in the same plot.
```
JamesWrigley, line 907: Minor thing, I would recommend making screenshots of the plots these generates and linking them in the docstring to give the users an idea of what to expect.

JamesWrigley (changes requested):
I didn't see anything critical, I think it's ok to address my comments later if you'd like to merge it now.

src/extra/recipes/base.py
daniloefl: We actually discussed it and I tried doing it right now, but the issue is that I have elements inside of the classes that are dctionaries in `CookieboxCalib`, for example. I would need somehow to flatten the structure of those dicts to be able to use xarray.

src/extra/recipes/cookiebox.py
daniloefl: God damn it! You are right. It is not the same, but it is close enough. I replaced it with your suggestion.

src/extra/recipes/cookiebox.py
daniloefl: This is giving it some slack, because we don't know how much the peak that I find is off. This is the automatic detection bit, which can fail, because the data is so noisy. If it fails, I raise an exception and write down that the user should try to figure out the parameters by themselves.

src/extra/recipes/cookiebox.py
daniloefl: Agreed!

src/extra/recipes/cookiebox.py
daniloefl: Agreed.

src/extra/recipes/cookiebox.py
daniloefl: Your suggestion was my original version and I agree it is simpler ... It has been proposed that I could find a scheme to avoid redoing it all. In practice, it is indeed annoying to re-build the full object if you are optimizing parameters (when the autodetection does not work).

src/extra/recipes/cookiebox.py
daniloefl: Agreed.

src/extra/recipes/grating.py
daniloefl: You are right ... I moved them to `setup()` and removed the default argument.

src/extra/recipes/cookiebox.py
daniloefl: This was "debugging math". You are right. Removed.

src/extra/recipes/cookiebox.py
daniloefl: I might do this later. I am not sure if these plots are ok to share (it is not commissioning data). It is nothing of consequence, but it might be worth checking.

src/extra/recipes/cookiebox.py
JamesWrigley: Ok, that seems fine but I would add a comment explaining it.

daniloefl:
## Changes done since the PR creation

As suggested by @tmichela and @TimBerberich:
- I created a `offline_tests` directory with integrated tests. As the runs cannot be put in GitHub, they are not run automatically as part of the CI/CD, but this could allow one to run them offline for tests.
- I switched towards using `logging`.
- I split the constructors into `__init__` and `setup`, such that `setup` does the actual hard work and `__init__` can always be called without arguments. In this way, I can instantiate the class for loading files from disk, and use `setup` if I want to make the fit from scratch.
- The name convention was changed: `CookieboxCalib` became `CookieboxCalibration` and `GratingCalib` became `Grating2DCalibration`. This opens the door for a future `Grating1DCalibration`.
- @tmichela s not hapy with ingredients, so it has been renamed `recipes`. I agree it makes more sense.
- Many docstrings have been added and many comments removed.
- Added missing type annotations.
- The `save` and `load` methods have been renamed `to_file` and `from_file` and now are part of the `BaseCalibration` class, which does most of the fiddling automatically. @JamesWrigley and @tmichela had proposed using `xarray`, but there are properties which are dicts of dicts, which would not work well with xarray, unless we flatten it all similarly to what I currently do. I am also not particularly happy with this, but I am open to suggestions.

src/extra/recipes/grating.py
JamesWrigley: Well, if there is only one of them for the foreseeable future then I don't think there's anything wrong with leaving them as default arguments :) But if more exist/are expected then it might be worth trying to automatically find them based on the device class or something, like the XGM and digitizer components.

JamesWrigley (approved):
I feel that the `CookieboxCalibration` component could be simplified in the future but so as not to hold this up, LGTM!

BTW, please either rebase the branch before merging or do a squash merge to keep the commit history on `master` clean-ish.

philsmt:
A few points that we should think about now, as they'd be hard to change later without breaking API (which we *usually* don't want to do):

*  Even more so than with `extra.components` I wonder whether it's a good idea to import all recipes in `extra.recipes.__init__` - just the Cookiebox one here already imports countless of dependencies, and if we double down on the recipes pattern we could end up with many different recipes importing various data science libraries. You may only want one, but cause the import of half the stack
* The class still expects source names, which ironically is something we actually discourage now with components and aliases. This is mostly fine though as long as the current interface is at least forward-compatible to only pass components and aliases in the future. A particular problem here could be the `tof_settings` parameter.
* EXtra-data and EXtra (components) aim to follow a stateless pattern, where any object change incurs a new object. `CookieboxCalibration` is quite drastically the opposite as a rich stateful object with automatic recomputation on changes. I see this pattern being useful in explorative analysis when tinkering around, which may be our target market here. But we probably want to stick with it for recipes in the future, as mixing patterns is a bad experience.
* Related to the point above, are any of these smart setters expensive? We generally also avoid having expensive property access - this can code behaviour quite unintuitive. It's not uncommon to carelessly set properties in a loop, which may blow up here. Something that's *actually* doing work should be a method. But of course we could keep the stateful pattern while simply making the setter's explicit. 
* As James also noted, I'm not a fan of the arbitrary HDF format. When we offer serialization in our tools, we should strive for more structured formats. One could change this in the future, and I would personally not even regret breaking compatibility, or only keep read support.
* It's quite bold to define such a generic interface like `BaseCalibration` when we're only starting out. All it really seems to add though is being a mixin for serialization. Why not make it just that?

daniloefl:
That is a long comment. Let me go point by point.

> A few points that we should think about now, as they'd be hard to change later without breaking API (which we _usually_ don't want to do):
> 
>     * Even more so than with `extra.components` I wonder whether it's a good idea to import all recipes in `extra.recipes.__init__` - just the Cookiebox one here already imports countless of dependencies, and if we double down on the recipes pattern we could end up with many different recipes importing various data science libraries. You may only want one, but cause the import of half the stack

I agree on this topic. I removed the imports in `__init__.py`.


> 
>     * The class still expects source names, which ironically is something we actually discourage now with components and aliases. This is mostly fine though as long as the current interface is at least forward-compatible to only pass components and aliases in the future. A particular problem here could be the `tof_settings` parameter.

Currently, the `tof_settings` also accepts `AdqRawChannel` (although this has not been fully tested). I see your point of keeping the usage of other objects, but part of the goal here is to reduce the work in doing this standard analysis. Compare the following two bits of code:
```
digi = 'SQS_DIGITIZER_UTC4/ADC/1:network'
tof_settings = {
       0:(digi, "1_A"),
       1:(digi, "1_C"),
       }
```

And:
```
digi = 'SQS_DIGITIZER_UTC4/ADC/1:network'
tof_settings = {
       0: AdqRawChannel(run,
                                      "1_A",
                                      digitizer=digi,
                                      first_pulse_offset=first_pulse_offset,
                                      single_pulse_length=single_pulse_length,
                                      interleaved=True, # necessary for a union of runs
                                      )
       1: AdqRawChannel(run,
                                      "1_C",
                                      digitizer=digi,
                                      first_pulse_offset=first_pulse_offset,
                                      single_pulse_length=single_pulse_length,
                                      interleaved=True, # necessary for a union of runs
                                      ),
       }
```

The work the user would have to do is much greater in the latter case. I would not aim for the latter only to keep things consistent if the former is more convenient, given that the goal here is convenience.

Additionally, I need to rebuild the object when applying this to a new run. It would be nice if the extra components had a way of outputting their state in a dictionary that I could simply save and then I could create a new element with the same settings for the new run. Of course there is `__dict__`, but how do I distinguish properties particular to the old run from properties that are valid when building a new object? I do need to keep several parameters the same between the initial setup and later calibration stages, otherwise the whole thing breaks down.

I do agree that it makes sense to change the energy source and the XGM source to be simply the corresponding EXtra objects. The issue in the three cases though, is that we need to match data from each source and that would ideally be done using `run.select(*all_sources, require_all=True)`. I could do that internally with the names of the sources, but using the EXtra objects, I have to update internal object elements. I have done that now by matching train IDs, but it is far more complicated than needed only to keep this interface. One has to start wondering whether this strict requirement is needed for the `recipes` in general.

> 
>     * EXtra-data and EXtra (components) aim to follow a stateless pattern, where any object change incurs a new object. `CookieboxCalibration` is quite drastically the opposite as a rich stateful object with automatic recomputation on changes. I see this pattern being useful in explorative analysis when tinkering around, which may be our target market here. But we probably want to stick with it for recipes in the future, as mixing patterns is a bad experience.
> 

I think for this kind of analysis and the future ones I have in mind, this pattern really is hard to avoid. We need to match several sources and changes in one may change others. I offer to remove the setters, to keep those properties hidden, but there would be a penalty in usability.

>     * Related to the point above, are any of these smart setters expensive? We generally also avoid having expensive property access - this can code behaviour quite unintuitive. It's not uncommon to carelessly set properties in a loop, which may blow up here. Something that's _actually_ doing work should be a method. But of course we could keep the stateful pattern while simply making the setter's explicit.

Agreed. I have changed the setters to proper methods.

> 
>     * As James also noted, I'm not a fan of the arbitrary HDF format. When we offer serialization in our tools, we should strive for more structured formats. One could change this in the future, and I would personally not even regret breaking compatibility, or only keep read support.

As I mentioned to James, the `xarray` Dataset format does not, to my knowledge, handle structured dicts or dicts of dicts, as I have here. Pickle has terrible dependencies on the package version and it is not great if we want to store some relatively large reference data used for the calibration, as diagnostics in the future (sometimes a few MB, in some long runs, GBs). I could indeed switch to `xarray`, but I see no benefit, as I would have to use my serialization code anyway to handle the dicts (but do tell me if you know an alternative) and everything else would be mostly the same.

> 
>     * It's quite bold to define such a generic interface like `BaseCalibration` when we're only starting out. All it really seems to add though is being a mixin for serialization. Why not make it just that?

I agree with that point. I changed it to be a Mixin.

daniloefl:
Following a discussion with @philsmt, I have done lazy commits instead for the non-standard packages to allow for the global imports in `__init__.py`. I had forgotten to adapt the API from `Grating2DCalibration` to also follow the pattern of receiving `KeyData` objects instead of the source and key names.

daniloefl:
The tests mentioned in https://github.com/European-XFEL/EXtra/pull/284#issuecomment-2627818988 have been updated.

PR #303, opened 2025-03-14
Fix support for the `acquisition_time` with newer Karabacon versions

JamesWrigley:
Before it would grab the entire vector rather than the first value in continuous scans. This will need some more tweaking in the future to select the first N values (where N is the number of steps) for non-continuous scans.


tmichela (approved):
LGTM

PR #304, opened 2025-03-14
Added auxiliary tool to calculate the TOF instrument response (optionally deconvolve), and Virtual Spectrometer light

daniloefl:
Two methods are added here with the purpose of improving the data quality of eTOF data:

- `TOFResponse` encodes the response function of the eTOFs and can be used to deconvolve and denoise it using Total-Variation-deconvolution. This acts on the raw time-of-flight trace. It is slow and requires tuning of a parameter $\lambda$, but it works with existing processes for calibrating the eTOF trace.
- `VSLight` correlates monochromator energy values with the eTOF monochromated traces to provide an automated calibration and deconvolution. It can be applied much faster than `TOFResponse`, has no parameter tuning, but requires continuous sweeps of the monochromator as a calibration run.

With the aim to include the `TOFResponse` as part of the existing process of the eToF calibration, the `CookieboxCalibration` object had its function `apply` split in two bits:
- `CookieboxCalibration.load_trace` simply loads and slices the trace for a new run, returning a `DataArray`.
- `CookieboxCalibration.calibrate` takes the result of `load_trace` and applies the effective calibration.

`CookieboxCalibration.apply` still exists and simply calls the two methods above.

The reason for the need of `load_trace` is that the data fetching step must be done in a way that is fully compatible with the setup of `CookieboxCalibration`, otherwise the calibration results are nonsensical. It therefore provides a way for the user to reapply the data fetching under the same settings as used for the calibration run.

With this separation of tasks, one could do, for instance:
```
my_new_run = open_run(...)
calibrator = CookieboxCalibration.from_file("my_calibration_constants.h5")
response_tof_4 = TOFResponse.from_file("my_instrumental_response_tof_4.h5")
# just get the data with the same settings as in calibration
trace = calibrator.load_trace(my_new_run)
# optionally do any processing step, such as deconvolving the spectra for each eTOF
trace.loc[dict(tof=4)] = response_tof_4.apply(trace)
# finally, calibrate
spectrum = calibrator.calibrate(trace)
```

Additionally, the `AdqRawChannel` has an `extra_cm_period` argument, which allows for extra common mode corrections to be applied to subtract noise.


src/extra/recipes/cookiebox.py
```diff
@@ -1019,95 +984,148 @@ def plot_fit(self, tof_id: int):
         plt.title(f"TOF {tof_id}")
         plt.show()
 
-    def apply(self, run: DataCollection) -> xr.Dataset:
+    def load_trace(self, run: DataCollection, **extra_kwargs_adq: Dict[str, Any]) -> xr.Dataset:
```
JamesWrigley, line 987: I'm not entirely sold on the idea of basically wrapping a constructor :thinking: Is it feasible to allow passing in a `DataArray` or `SourceData` object for the trace instead, in case one already has those lying around in a script? Or is it important that that e.g. a `DataArray` with the pulse information was constructed from an `AdqRawChannel` with specific settings?

src/extra/recipes/cookiebox.py
```diff
@@ -1019,95 +984,148 @@ def plot_fit(self, tof_id: int):
...
+        The method `apply` provides a direct way to retrieve the trace and calbrate it.
+        The methods `load_trace` and `calibrate` allow one to apply an intermediate processing
+        step between them.
+
+        Args:
+          trace: A pre-processed trace retrieved with the *same* `AdqRawChannel` settings as this calibration object.
+                 Its axes are expected to be ('trainId', 'pulseIndex', 'sample', 'tof').
+                 It is recommended to use always `load_trace` to obtain this.
+
+        Returns: the calibrated data as an xarray DataArray. The DataArray obs cotains the calibrated data.
```
JamesWrigley, line 1043: Nitpicking:
```suggestion
        Returns: the calibrated data as an xarray DataArray.
```

src/extra/recipes/cookiebox_deconvolve.py
```diff
@@ -0,0 +1,880 @@
...
+    """
+    Given a run with the Cookiebox in counting mode, obtain its impulse response.
+
+    Example usage:
+    ```
+    tof_id = 4
+    channel = "3_A"
+    digitizer = "SQS_DIGITIZER_UTC4/ADC/1:network"
+    calib_run = open_run(proposal=900485, run=320)
+    calib_run = calib_run.select([digitizer, mpod_source], require_all=True)
```
JamesWrigley, line 337: Should we define `mpod_source` or will that be known by the users?

src/extra/recipes/vs.py
```diff
@@ -0,0 +1,890 @@
...
+    peak_idx = np.argmax(smoothened)
+    return peak_idx - 200
+
+def fwhm(x: np.ndarray, y: np.ndarray) -> float:
+    """Return the full width at half maximum of x."""
+    # half maximum
+    half_max = 0.5*np.amax(y)
+    idx = np.where(y > half_max)[0]
+    if len(idx) < 2:
+        return 999
```
JamesWrigley, line 43: Would maybe 0 or -1 be a more obvious signal? Or even throwing an exception? 999 is a plausible (if unlikely) FWHM.

src/extra/recipes/vs.py
```diff
@@ -0,0 +1,890 @@
...
+        fig = plt.figure(figsize=(10, 8))
+        if tof_ids is None:
+            tof_ids = self.resolution.keys()
+        for tof_id in tof_ids:
+            plt.errorbar(self.e_probe, [self.resolution[tof_id][i]['fwhm'] for i, _ in enumerate(self.resolution[tof_id].keys())],
+                         xerr=self.e_width/2, alpha=0.5,
+                         fmt='o',
+                         lw=2, label=f"eTOF {tof_id}")
+        plt.xlabel("Energy [eV]")
+        plt.ylabel("Resolutison [eV]")
```
JamesWrigley, line 624: Nitpicking:
```suggestion
        plt.ylabel("Resolution [eV]")
```

src/extra/recipes/vs_utils.py
```diff
@@ -0,0 +1,245 @@
...
+
+        Args:
+          X: {array-like, sparse matrix} of shape (n_samples, n_features)
+            The input data.
+
+        Returns: {array-like, sparse matrix} of shape (n_samples, n_outputs)
+            Multi-output targets predicted across multiple predictors.
+            Note: Separate models are generated for each predictor.
+        """
+        from joblib import Parallel, delayed
```
JamesWrigley, line 236: `joblib` needs to be added to `pyproject.toml` if we're going to use it. But why use joblib over e.g. `multiprocessing.Pool.map()`?

JamesWrigley (changes requested):
Mostly looks ok to me, but I'm not too familiar with this field. Some general comments:
- In some places there are commented out code snippets. I think some of them might be pseudocode for documentation, but other snippets look like they could be deleted.
- In the docstrings there's lots of examples in code blocks (very cool :+1:), if you begin them with '```python' then mkdocs will use Python syntax highlighting for them.
- I know it's difficult to test the output, but this should have at least some smoke tests using dummy data.
- The number of serializeable things that can produce intermediate files seems to be increasing :eyes: I wonder if we should try to introduce some structure to the file location/naming. Something for later though.

docs/recipes/vs.md
```diff
@@ -0,0 +1,3 @@
+# Virtual diagnostics
+
+::: extra.recipes.VSLight
```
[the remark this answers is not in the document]
daniloefl: Agreed. Thank you!

src/extra/components/adq.py
```diff
@@ -108,7 +110,8 @@ class AdqRawChannel:
     def __init__(self, data, channel, digitizer=None, pulses=None,
                  interleaved=None, clock_ratio=None, sample_dim='sample',
                  first_pulse_offset=10000, single_pulse_length=25000,
-                 cm_period=None, baselevel=None, baseline=np.s_[:1000]):
+                 cm_period=None, baselevel=None, baseline=np.s_[:1000],
+                 extra_cm_period = list()):
```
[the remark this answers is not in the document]
daniloefl: Isn't this a problem, though? According to ... https://docs.python-guide.org/writing/gotchas/

I don't do that, though, but could be a concern if there are changes, or not?

src/extra/components/adq.py
```diff
@@ -348,7 +355,7 @@ def _preprocess(self, data, out):
 
         if self._cm_period > 0:
             # Apply common mode corrections (includes baselevel).
-            self._correct_cm_by_train(data, out, self._cm_period,
+            self._correct_cm_by_train(data, out, [self._cm_period] + list(self._extra_cm_period),
```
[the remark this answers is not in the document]
daniloefl: If it is sent as a `np.array`, the consequences are a disaster (I have heard ... of course I wouldn't have done it myself and have learned this from experience, would I).

src/extra/recipes/cookiebox.py
daniloefl: It is *extremely* important that the `AdqRawChannel` used for the `calibrate` step has exactly the same settings (with very few exceptions). This is why I split it in `load_trace` and `calibrate`, so that people can do it themselves and then call `calibrate` only, but at their own peril. Still, in most cases you would not want to do it yourself and you would want to use the same settings. I also don't like that I have to call the constructor myself, but given the constraints, are there alternatives?

src/extra/recipes/cookiebox_deconvolve.py
daniloefl: They would not and this should not be here: copy-paste mistake. Good catch!

src/extra/recipes/vs.py
daniloefl: Agreed. I don't want to throw an exception, as this is for validation only, but using -1 is a good idea.

daniloefl:
Thanks, James! I agree with the comments on the docstrings. I have already looked through them, but I will check them again to discard old code.
It is extremely hard to test with dummy data though. All of the code so far in the recipes require very particular data, which we cannot access without GPFS. What I did before was write offline tests, which could be run by a person before committing.

src/extra/components/adq.py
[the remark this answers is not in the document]
JamesWrigley: Ah, TIL. NVM then, `list()` is definitely better.

src/extra/recipes/cookiebox.py
JamesWrigley: Ok, if it's that important then I think it's unavoidable. I also can't think of a better way.

src/extra/recipes/vs_utils.py
daniloefl: I agree. Done.

JamesWrigley: I'm a bit confused, it looks like joblib is still being used?

JamesWrigley: Ah now I see, it was replaced in `fit()` but not `predict()`. Is it feasible to remove from `predict()` too?

daniloefl: I'm just a twat, who did not notice it is also used elsewhere. I pushed the update.

JamesWrigley (changes requested):

docs/recipes/index.md
```diff
@@ -8,3 +8,8 @@ Recipes index:
     - [CookieboxCalibration][extra.recipes.CookieboxCalibration]
 - [Grating](grating.md)
     - [Grating2DCalibration][extra.recipes.Grating2DCalibration]
+- [Virtual Spectrometer](vs.md)
+    - [VSLight][extra.recipes.VSLight]
+- [eTOF response analysis](cookiebox_deconvolve.md)
```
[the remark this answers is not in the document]
daniloefl: Thank you! I had it, but did not push. I also removed the `Grating1DCalibration`, which was broken, since it does not exist yet (I also have that code ready, but I am waiting to push it in a separate PR, to avoid chaos).

JamesWrigley (approved):
LGTM! But please either rebase the branch or do a squash merge to keep the commit history clean.

daniloefl:
I merged with the `master` branch. Thank you. I will merge with squash commits anyway.

PR #310, opened 2025-03-31
Fix pulse alignment in ADQ raw data component

philsmt:
Speaking so much of (train) alignment issues last Friday, there were actually some lurking in the ADQ raw data component. 

This PR fixes two independent issues separated by commits:

1. fc80996c80177030b0f6d6e2e8f2b4e8cb827152: the component did not actually align the pulse information it uses internally to the actual trains for which the ADQ saved data for. This did not actually cause issues in most circumstances, as the pulse data is associated via train IDs, but could cause the `pasha` kernel in `.pulse_edges()` to break. For consistency, ADQ data is now always aligned before, and the pulse information selected properly.

2. 163f63b4cc1f3cdace98d0a1e0aece029a0315fc: This was a quite tricky bug in practice, as it only occured for variable-pulse runs. The `._prepare_pulses()` generates a dataframe with data about the pulse pattern per train including the first and last index in a global pulse list:

    |  | `first` | `last` |
    | - | - | - |
    | $0$ | $0$ | $N_0$ |
    | $1$ | $N_0$ | $N_0+N_1$ |
    | $2$ | $N_0+N_1$ | $N_0+N_1+N_2$ |
    | $3$ | $N_0+N_1+N_2$ | $N_0+N_1+N_2+N_3$ |
    | $m$ | $\sum \limits_{i=0}^{m-1} N_i$ | $\sum \limits_{i=0}^{m} N_i$ |

    The original implementation however gave this:

    |  | `first` | `last` |
    | - | - | - |
    | $0$ | $0$ | $N_0$ |
    | $1$ | $N_1$ | $2 N_1$ |
    | $2$ | $N_1+N_2$ | $N_1+2 N_2$ |
    | $3$ | $N_1+N_2+N_3$ | $N_1+N_2+2 N_3$ |
    | $m$ | $\sum \limits_{i=1}^{m} N_i$ | $\sum \limits_{i=1}^{m} N_i + N_m$ |

    This is naturally identical for $N_i = {\rm const}$, and might even go unnoticed for small variations in pulse count. In the past experiment however, we were switching between 2 and 92 pulses, which sooner or later caused the code to break by jumping train boundaries.

### DIFF
--- a/src/extra/components/adq.py
+++ b/src/extra/components/adq.py
@@ -411,7 +411,7 @@ class AdqRawChannel:
 
         if pulses_end > data.shape[-1]:
             raise ValueError(f'trace axis too short for {num_pulses} pulses '
-                             f'located at {pulses_start}:{pulses_end}]')
+                             f'located at {pulses_start}:{pulses_end}')
 
         return data[..., pulses_start:pulses_end].reshape(
             *data.shape[:-1], num_pulses, samples_per_pulse)
--- a/src/extra/components/adq.py
+++ b/src/extra/components/adq.py
@@ -390,8 +390,8 @@ class AdqRawChannel:
 
         # Generate offsets of first pulse and last pulse of each
         # train relative to all pulses.
-        pulse_first = num_pulses.cumsum() - num_pulses.iloc[0]
-        pulse_last = pulse_first + num_pulses
+        pulse_last = num_pulses.cumsum()
+        pulse_first = pulse_last - num_pulses
 
         # Combine pulse layout into a single dataframe.
         # TODO: samples_per_pulses is currently assumed to be constant
--- a/src/extra/components/adq.py
+++ b/src/extra/components/adq.py
@@ -356,23 +356,31 @@ class AdqRawChannel:
 
         return out
 
-    def _prepare_pulses(self):
+    def _prepare_pulses(self, train_ids):
         """Prepare pulse information."""
 
         if self._pulses is None:
             raise RuntimeError('component must be initialized with pulse '
                                'information for this operation')
 
-        pulse_ids = self._pulses.get_pulse_ids(labelled=True)
+        pulse_ids = self._pulses.pulse_ids(labelled=True)
         pids_by_train = pulse_ids.groupby(level=0)
 
         # Number of pulses per train.
         num_pulses = pids_by_train.count()
 
+        # Align pulses to passed train IDs of actual data.
+        try:
+            num_pulses = num_pulses.loc[train_ids]
+        except KeyError:
+            raise ValueError('missing pulse information for one or more '
+                             'trains') from None
+
         # Samples per pulse based on the shortest difference between
         # pulses if available. All code below using this value is
         # protected against out-of-bounds access.
         try:
+            # Beware, pulse_period is not aligned to train_ids here!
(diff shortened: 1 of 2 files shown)


src/extra/components/adq.py
```diff
@@ -356,23 +356,31 @@ def _preprocess(self, data, out):
...
 
         # Number of pulses per train.
         num_pulses = pids_by_train.count()
 
+        # Align pulses to passed train IDs of actual data.
+        try:
+            num_pulses = num_pulses.loc[train_ids]
+        except KeyError:
+            raise ValueError('missing pulse information for one or more '
+                             'trains') from None
```
takluyver, line 377: The bunch pattern table comes as instrument data, so it could in principle have gaps. Hopefully that's very rare, though, so I think it's OK to do this and see if it occurs in practice.

philsmt: Yes, this indeed happens and the *correct* behaviour in the future would be to try and interpolate such events. As it happens `PumpProbePulses` already does some limited interpolations (which can be switched off). But that's the place I would most likely prefer such a feature.

Raising this exception here however is not a change in behaviour, as before the code would've simply broken in more cryptic ways.

takluyver: I'd say if it comes up, the correct behaviour is more likely to drop the corresponding digitizer data where we don't know the number of pulses, rather than trying to interpolate it. I imagine it's possible to have e.g. an alternating train pattern, so the number before & after the gap are the same, but the missing value is different.

takluyver: (But it's fine to raise an exception for now, either way)

src/extra/components/adq.py
```diff
@@ -1285,15 +1293,18 @@ def pulse_edge_array(self, labelled=True, squeeze_edges=True,
...
-        pulse_ids, pulse_layout = self._prepare_pulses()
+        pulse_ids, pulse_layout = self._prepare_pulses(raw_key.train_ids)
 
         if max_edges is None:
             max_edges = pulse_layout['length'].max() // 100
 
         # Set-up pasha for parallel processing.
         psh = self._prepare_pasha(parallel)
-        trace_out = np.zeros(self._raw_key.entry_shape, dtype=np.float32)
+        trace_out = np.zeros(raw_key.entry_shape, dtype=np.float32)
```
daniloefl, line 1307: Silly thing, but you assume `float32`. Could you not refer to another `dtype`, so that you keep the same type as the input/

src/extra/components/adq.py
```diff
@@ -356,23 +356,31 @@ def _preprocess(self, data, out):
 
         return out
 
-    def _prepare_pulses(self):
+    def _prepare_pulses(self, train_ids):
```
daniloefl, line 359: Why not have train_ids by default None and revert to the old behaviour to make it backwards compatible?

daniloefl:
Other than silly things, it also LGTM. Feel free to ignore my comments, if this is urgent.

src/extra/components/adq.py
philsmt: This is a private method, and I've fixed all its call sites.

philsmt:
Thanks for the quick review!

Indeed I anticipated creating new tests, but there's not any data test yet 😬

PR #312, opened 2025-04-03
Use _reshape_flat_pulses rather than _reshape_pulses_by_train for pulse splitting

philsmt:
There are methods in `AdqRawChannel` to split train traces into pulse traces, via simple reshaping and through a Cython function. The latter exists mostly to allow for flat pulse axes with non-constant pulse counts, but can also handle train traces that are too short for the number of pulses present.

`AdqRawChannel.pulse_edges()` and `AdqRawChannel.pulse_edge_array()` used the former method so far, which ran into the trace length problem for a couple of runs. Hence, this PR changes them to use the latter method instead. There is an insignificant speed overhead between the two, which in most cases is entirely overshadowed by IO and edge discrimination.

### DIFF
--- a/src/extra/components/adq.py
+++ b/src/extra/components/adq.py
@@ -1302,23 +1302,36 @@ class AdqRawChannel:
         if max_edges is None:
             max_edges = pulse_layout['length'].max() // 100
 
+        # Convert to unlabelled array for use in native code later.
+        pulse_ids = pulse_ids.to_numpy()
+
         # Set-up pasha for parallel processing.
         psh = self._prepare_pasha(parallel)
-        trace_out = np.zeros(raw_key.entry_shape, dtype=np.float32)
         edges = psh.alloc(shape=(len(pulse_ids), max_edges),
                           dtype=np.float32, fill=np.nan)
         amplitudes = psh.alloc(like=edges, fill=np.nan)
 
+        # These buffers are only used locally and not in shared memory!
+        trace_out = np.zeros(raw_key.entry_shape, dtype=np.float32)
+        trace_split = np.zeros(
+            (pulse_layout['count'].max(), pulse_layout['length'].max()),
+            dtype=np.float32)
+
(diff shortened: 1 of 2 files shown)


src/extra/components/adq.py
```diff
@@ -1302,23 +1302,32 @@ def pulse_edge_array(self, labelled=True, squeeze_edges=True,
...
+            self._reshape_flat_pulses(
+                self._preprocess(trace_in, trace_out)[np.newaxis, :],
+                trace_split, pulse_ids[first:last], pulse_len)
 
-            for pulse_trace, pulse_index in zip(pulse_traces, pulse_sel):
-                edge_func(signal=pulse_trace, edges=edges[pulse_index],
+            # No need to slice trace_split to the actual number of
+            # pulses, as first:last limits the loop
+            for signal, pulse_index in zip(trace_split, range(first, last)):
+                edge_func(signal=signal[:pulse_len], edges=edges[pulse_index],
```
daniloefl, line 1330: Given that only `pulse_len` samples are `memcpy`ed in `trace_split`, why slice `signal[:pulse_len]`. It is not a problem, but there are a lot of details in the slicing that happens in auxiliary functions. It might be helpful to add some comments on what is going on there for future maintainability.

daniloefl:
LGTM. I cannot see anything wrong with it, except that it might benefit from some more comments, because several variables are modified inside Cython functions or external functions and this is not so easy to follow, thinking about future development.

src/extra/components/adq.py
philsmt: Good catch, thanks!

philsmt: Ahh, on second thought this doesn't quite work like that unless we'd zero out the buffer on every train. Imagine a train with longer pulses followed by a train with short ones - we'd leave some samples at the end of the buffer untouched, which the edge function would see.

PR #313, opened 2025-04-03
Account for additional indices in PulsePattern.is_constant_pattern

philsmt:
Thomas noticed that `PulsePattern.is_constant_pattern` does not take the additional flags introduced by the `PumpProbePulses` multi index into account, specifically which pulses are pumped and unpumped. As such, an alternating train pattern was incorrectly described as *constant*.

This PR extends the method in an hopefully forwards-compatible way by shifting any additional indices into columns of a dataframe, and check the constant condition for all of them.

Creating this for testing purposes immediately, will follow up with tests later this week.


philsmt:
Added tests, this is ready for review!

takluyver:
LGTM, thanks!

philsmt:
Thanks for review

PR #316, opened 2025-04-09
Add tests to ADQ component for actual data retrieval

philsmt:
I was trying to write too perfect tests, and it took forever. With some recent issues in mind, I finished some *solid* tests that cover a good deal.

As it always is, I found a couple of bugs along the way.

### DIFF
--- a/src/extra/components/adq.py
+++ b/src/extra/components/adq.py
@@ -374,7 +374,10 @@ class AdqRawChannel:
         pids_by_train = pulse_ids.groupby(level=0)
 
         # Number of pulses per train.
-        num_pulses = pids_by_train.count()
+        # Note that using .pulse_counts() is different than
+        # pids_by_train.count(), as the latter may miss trains due to
+        # having no pulses while still having the record of that.
+        num_pulses = self._pulses.pulse_counts()
 
         # Align pulses to passed train IDs of actual data.
         try:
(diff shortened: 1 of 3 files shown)


src/extra/components/adq.py
```diff
@@ -367,7 +367,10 @@ def _prepare_pulses(self, train_ids):
         pids_by_train = pulse_ids.groupby(level=0)
 
         # Number of pulses per train.
-        num_pulses = pids_by_train.count()
+        # Note that using .pulse_count() is different than
```
takluyver, line 370:
```suggestion
        # Note that using .pulse_counts() is different than
```

tests/test_components_adq.py
```diff
@@ -174,3 +176,129 @@ def test_pull_baseline(mock_sqs_remi_run, in_dtype):
...
+    assert isinstance(np_data, np.ndarray)
+
+    for data in [xr_data.data, np_data]:
+        assert data.shape == (100, 30000)
+        assert data.dtype == np.float32
+
+        assert np.isclose(data[0, 0], 0.0)
+        assert np.isclose(data.min(), 0.0)
+        assert np.isclose(data.max(), 5.0)
+        assert np.all(data.argmax(axis=1) == 8974)
```
takluyver, line 200: The `numpy.testing` module has various helpers that can give more detailed information if the assertion fails - in particular `assert_array_equal` and `assert_allclose`. They allow both array/array comparisons and array/scalar.

```suggestion
        np.testing.assert_array_equal(data.argmax(axis=1), 8974)
```

takluyver:
Other than that, LGTM :+1:

tests/test_components_adq.py
philsmt: Hmmm, I'm regularly using that, not sure why I didn't here 🤔

PR #317, opened 2025-04-10
Add PumpProbePulses.pumped_pulses_ratios()

philsmt:
I recently needed to frequently distinguish regular trains with all or most pulses pumped vs trains where intentionally only few or no pulses are pumped for reference. This was way more work than I wanted to, mostly because of necessary train alignment between these modes.

This PR adds a corresponding method `.pumped_pulses_ratios()` returning such a series automatically.

@takluyver This could also be used to distinguish pump-probe patterns

### DIFF
--- a/src/extra/components/pulses.py
+++ b/src/extra/components/pulses.py
@@ -2248,3 +2248,12 @@ class DldPulses(PulsePattern):
         warn("Use triggers() instead of get_triggers()",
              DeprecationWarning, stacklevel=2)
         return self.triggers(*args, **kwargs)
+
+    @wraps(PumpProbePulses.pumped_pulses_ratios)
+    def pumped_pulses_ratios(self, *args, **kwargs):
+        pids = self.pulse_ids(copy=False)
+
+        if 'ppl' not in pids.index.names:
+            raise ValueError('only available with PPL information')
+
+        return PumpProbePulses.pumped_pulses_ratios(self, *args, **kwargs)
--- a/src/extra/components/pulses.py
+++ b/src/extra/components/pulses.py
@@ -1985,6 +1985,59 @@ class PumpProbePulses(XrayPulses, OpticalLaserPulses):
         else:
             raise ValueError(f"{field=!r} parameter was not 'fel'/'ppl'/None")
 
+    def pumped_pulses_ratios(self, ppl_only_value=np.nan, labelled=True):
+        """Determine ratio of pumped pulses per train.
+
+        Args:
+            ppl_only_value (float, optional): Value for trains only
+                containing PPL pulses, same value as for pulses without
+                any pulses (np.nan) by default.
+            labelled (bool, optional): Whether a labelled pandas Series
+                (default) or unlabelled numpy array is returned.
+
+        Returns:
+            (pandas.Series or numpy.ndarray): Number of pulses per
+                train, indexed by train ID if labelled is True.
+        """
+
+        pids = self.pulse_ids(copy=False)
+
+        try:
+            fel_count = pids[:, :, True, :].groupby('trainId').count()
+        except KeyError:
+            fel_count = pd.Series([])
+
+        try:
+            pumped_count = pids[:, :, True, True].groupby('trainId').count()
+        except KeyError:
+            pumped_count = pd.Series([])
+
+        # Compute the ratio for trains with at least one pumped pulse.
+        ratios = pumped_count / fel_count.loc[pumped_count.index]
+
+        # Extend the series to all expected trains, filling with NaN.
+        ratios = self._extend_all_trains(ratios, np.nan)
+
+        # Set those trains with all unpumped pulses to 0.0.
+        fel_only_index = fel_count.index.difference(pumped_count.index)
+        ratios.loc[fel_only_index] = 0.0
+
+        # Set those trains with no FEL pulses to the desired fill value.
+        if ppl_only_value is not np.nan:
+            # If one only cares for the index labels of a groupby,
+            # pd.SeriesGroupBy.count() is indeed faster than
+            # pd.SeriesGroupBy.groups, likely due additional objects
+            # created by the latter.
+            try:
+                ppl_only_index = pids[:, :, :, True].groupby('trainId') \
+                    .count().index.difference(fel_count.index)
+            except KeyError:
+                pass
+            else:
+                ratios.loc[ppl_only_index] = ppl_only_value
+
+        return ratios if labelled else ratios.to_numpy()
+
 
 class ManualPulses(PulsePattern):
     """Manual pulse patterns.
--- a/tests/test_components_pulses.py
+++ b/tests/test_components_pulses.py
@@ -470,8 +470,8 @@ def test_optical_laser_specials(mock_spb_aux_run):
     assert (pulses.pulse_counts() == 0).all()
 
     # Different laser seed by string.
-    pulses = OpticalLaserPulses(run, ppl_seed='MID')
-    assert pulses.ppl_seed == PPL_BITS.LP_SASE2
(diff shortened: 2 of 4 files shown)


src/extra/components/pulses.py
```diff
@@ -1507,6 +1507,59 @@ def pulse_mask(self, labelled=True, field=None):
         else:
             raise ValueError(f"{field=!r} parameter was not 'fel'/'ppl'/None")
 
+    def pumped_pulses_ratios(self, ppl_only_value=np.nan, labelled=True):
```
takluyver, line 1510: `labelled` is never used. I don't have a strong preference whether we implement it or remove it.

tests/test_components_pulses.py
```diff
@@ -740,3 +740,27 @@ def test_pump_probe_specials(mock_spb_aux_run, mock_sqs_remi_run):
...
+    pulses._pulse_ids = pd.Series(
+        [300, 310, 300, 300, 300, 310],
+        index=pd.MultiIndex.from_tuples([
+            (1000, 0, True, True),
+            (1000, 0, True, False),
+            (1001, 0, True, False),
+            (1002, 0, False, True),
+            (1003, 0, True, True),
+            (1003, 0, True, True),
+        ], names=['trainId', 'pulseIndex', 'fel', 'ppl']))
```
takluyver, line 756: In general I think mocking out bits of the innards of a class like this to test its public API is annoyingly brittle, and it's better to make some suitable input. I won't hold the PR up over it, though. Maybe we need better facilities for making mock data.

takluyver:
I think doing this with a pandas series makes it more complex than using a 2D array. I haven't tested this, but as an idea:

```python
fel_count = self.pulse_mask(field='fel').sum(axis=1)
ppl_count = self.pulse_mask(field='ppl').sum(axis=1)
ratio = ppl_count / fel_count
ratio[fel_count = 0] = fill_value
```

Might want to avoid division by zero; `np.divide()` takes a `where=` parameter.

philsmt:
> I think doing this with a pandas series makes it more complex than using a 2D array. 

Assuming a constant number of pulses per train is no longer useful for me, in particular when pump-probe is used. That's why I based the `PulsePattern` family mostly on `pandas` types.

Going forward the more ubiquitous use of frame filters will likely make this even more common. I don't think we can get away with using linear pulse axes more often.

src/extra/components/pulses.py
philsmt: Well spotted. I realized at the end all other methods carry it, but forgot the implementation...

tests/test_components_pulses.py
philsmt: Generally I agree, though I found it acceptable for `PulsePattern` given the explicit caching mechanism.

As you guessed correctly, I wanted to avoid having to creating different mock devices with the current interface to test a very particular scenario. Ideally we would continue to be able to test without Maxwell, and we also cannot rely on runs sticking around forever unless we copy them to a defined place. Maybe some serialized data in the repository that is unpacked into EXDF?

takluyver:
I'd maybe add `.pulse_mask()` - or rather the variant with the `field=` parameter - to `DldPulses`, to match `PumpProbePulses`. It looks to me like the implementation of that plus `pumped_pulses_ratios()` using the mask arrays is still simpler than the MultiIndex version, and the enhanced pulse_mask could be useful as well. Up to you, though.

src/extra/components/pulses.py
```diff
@@ -1507,6 +1507,59 @@ def pulse_mask(self, labelled=True, field=None):
...
+        except KeyError:
+            fel_count = pd.Series([])
+
+        try:
+            pumped_count = pids[:, :, True, True].groupby('trainId').count()
+        except KeyError:
+            pumped_count = pd.Series([])
+
+        # Compute the ratio for trains with at least one pumped pulse.
+        ratios = pumped_count / fel_count.loc[pumped_count.index]
```
fadybishara, line 1538: I'm not sure I follow the logic here, if `pumped_count` and `fel_count` have different trains, how do you know that `fel_count` will have *more* trains? If not, wouldn't this throw an error? 

It's likely I misunderstood something but if not, perhaps a simple solution would be to do and explicit outer merge like
```py
joint_count = pd.concat([fel_count, pumped_count], keys=['fel', 'ppl'], axis=1, join='outer'
ratios = joint_counts.ppl.div(joint_counts.fel)
```

fadybishara: Oh, never mind, I get it -- all the trains have FEL pulses but not necessarily PPL pulses. Nevertheless, just because it should not happen doesn't mean it cannot happen, no? 

(Also, a better way to do what I suggested is with `np.intersect1d` on the train IDs -- but probably none of what I suggested is necessary.)

philsmt: Not quite - this line is not about FEL vs PPL pulses, but FEL vs FEL+PPL pulses. The set of trains with pumped pulses is a (not necessarily proper) subset of the set of trains with FEL pulses. It becomes clear when comparing line 1528 and line 1533, the latter makes a stricter indexing.

philsmt:
> I'd maybe add `.pulse_mask()` - or rather the variant with the `field=` parameter - to `DldPulses`, to match `PumpProbePulses`. It looks to me like the implementation of that plus `pumped_pulses_ratios()` using the mask arrays is still simpler than the MultiIndex version, and the enhanced pulse_mask could be useful as well. Up to you, though.

I had a thought about this, but this implementation would run into the problem of selected trains vs trains with data. As with `KeyData.data_counts()`, the statistics methods of these components strive to return a result for all selected trains, whether there is data or not.

src/extra/components/pulses.py
```diff
@@ -1507,6 +1507,59 @@ def pulse_mask(self, labelled=True, field=None):
...
+        fel_only_index = fel_count.index.difference(pumped_count.index)
+        ratios.loc[fel_only_index] = 0.0
+
+        # Set those trains with no FEL pulses to the desired fill value.
+        if ppl_only_value is not np.nan:
+            # If one only cares for the index labels of a groupby,
+            # pd.SeriesGroupBy.count() is indeed faster than
+            # pd.SeriesGroupBy.groups, likely due additional objects
+            # created by the latter.
+            try:
```
fadybishara, line 1553: Seems to me to be more complicated than necessary, why not do the following?
```py
try:
    ppl_only_index = pids[:, :, False, True].groupby('trainId').count()
except KeyError:
    ppl_only_index = pd.Series([])
```

philsmt:
Tryping to wrap this up, I noticed an additional point to consider for the discussion of going via pulse IDs or pulse mask during the plotting code recently. The former are cached in the object, while the latter may have to be generated and/or re-read. This can make a non-insignificant difference in runtime for large runs.

src/extra/components/pulses.py
philsmt: This may come very late, but I'm not quite sure from your last post whether there is any further need for discussion or changes here.

fadybishara: Nope, this is resolved.

src/extra/components/pulses.py
fadybishara: Hmm... I'm not sure what this is about any more 🙃, and I'm not even sure it's correct so I'll just resolve this comment as well.

philsmt: I think you meant to to the indexing with an empty series, rather than gate the indexing behind not throwing a `KeyError`. I think your version works, but I like the explicitness of handling it only in the `else` branch.

tests/test_components_pulses.py
philsmt: Refactore the test to use the mocked data from the timeserver device.

PR #318, opened 2025-04-10
Add calibration for 1D grating and improve existing interface in `recipes`

daniloefl:
- Calibrates 1D grating information (e.g.: collected from the Gotthard detector in SQS). Example of usage below.
- `CookieboxCalibration`: allow flagging of bad calibration points, to allow using others in the `mask_calibration_point(tof_id, energy)` function.
- `CookieboxCalibration`: Order `plot_calibration_data` y axis by energy to avoid unordered axis in case the energy scan was done out of monotonically increasing order.
- `CookieboxCalibration`: Use Auger-Meitner mean to align all plots in `plot_calibration_data` by the Auger-Meitner peak.
- `CookieboxCalibration`: New function `plot_calibrations`, which plots only the calibration factors in the same plot for diagnostics. Often this is all that one is interested in.
- `Grating2DCalibration`: Crop image based on rotation angle to avoid artifacts caused by out-of-bound effects. Add background root-mean-squared error as uncertainty band in the output.
- `Grating2DCalibration`: If provided, use extra motor information as independent variable in the fit (useful when data includes motor movements additionally).

Example usage of the 1D grating energy calibration:
```    
bkg_run = open_run(proposal=900485, run=590)    
calib_run = open_run(proposal=900485, run=611)    
scan = Scan(calib_run["SA3_XTD10_MONO/MDL/PHOTON_ENERGY", "actualEnergy"])    
grating_signal = calib_run["SQS_EXP_GH2-2/CORR/RECEIVER:daqOutput", "data.adc"]    
grating_bkg = bkg_run["SQS_EXP_GH2-2/CORR/RECEIVER:daqOutput", "data.adc"]    
pulses = XrayPulses(calib_run)    
grating_calib = Grating1DCalibration()    
grating_calib.setup(grating_signal, grating_bkg, scan, pulses)    

# apply it in new data    
new_run = open_run(...)    
grating_calib.apply(new_run)    
```


src/extra/recipes/cookiebox.py
```diff
@@ -1095,6 +1186,7 @@ def apply_correction(tof_id, tof_trace):
             #o = o - self.offset[tof_id][None, None, :]
             # apply Jacobian
             o = o*self.jacobian[tof_id][None, None, :]
+            o[np.isnan(o)] = 0.0
```
JamesWrigley, line 1189: Maybe it would be safer to use `~np.isfinite(o)`? I've used `isnan` in the past and inevitably been tripped up by infs. Could also use `np.nan_to_num(o, copy=False)`: https://numpy.org/doc/2.0/reference/generated/numpy.nan_to_num.html

src/extra/recipes/grating.py
```diff
@@ -109,10 +153,39 @@ def _fromdict(self, all_data):
...
+        Estimate difference between samples in neighbour pulses.
+
+        Args:
+          pulses: XrayPulses element for the run.
+
+        Returns: Sample difference between neighbour pulses.
+        """
+        pulse_ids = pulses.pulse_ids(labelled=True)
+        pids_by_train = pulse_ids.groupby(level=0)
+        pulse_period = int(pids_by_train.diff().min())
```
JamesWrigley, line 176: Out of curiosity, do I understand correctly that this is computing the period between pulses in units of entries in the bunch pattern table (or global pulse ID)?

src/extra/recipes/grating.py
```diff
@@ -122,45 +195,108 @@ def load_data(self):
...
+            self.bkg_unc = bkg_unc
+            self.bkg = np.zeros_like(data).mean(0)
+        self.calibration_unc = bkg_unc
+        # skip offset and collect pulse data each pulse_period samples only
+        self.calibration_data = data[:, self.offset::self.pulse_period, self.min_pixel:self.max_pixel]
+        self.calibration_unc = self.calibration_unc[self.offset::self.pulse_period, self.min_pixel:self.max_pixel]
+        # average over pulses
+        self.calibration_data = np.mean(self.calibration_data, axis=1)
+        self.calibration_unc = np.mean(self.calibration_unc, axis=0)
+        self.calibration_mask = np.array([True for _ in range(self.calibration_data.shape[0])])
```
JamesWrigley, line 216: Nitpicking, an alternative would be:
```suggestion
        self.calibration_mask = np.ones(self.calibration_data.shape[0], dtype=bool)
```

JamesWrigley (changes requested):

src/extra/recipes/cookiebox.py
```diff
@@ -726,21 +751,26 @@ def plot_calibration_data(self):
...
-                      aspect=(stop_roi - auger_start_roi)/n_energies,
-                      norm=Normalize(vmax=np.amax(temp))
+            temp = data[tof_id].copy()
+            ax.imshow(temp[idx, sample_auger:stop_roi],
+                      extent=[sample_auger, stop_roi, np.amin(energies), np.amax(energies)],
+                      aspect="auto",
+                      norm=Normalize(vmax=np.amax(temp)),
                      )
-            ax.set_yticklabels((energies+0.5).astype(int))
+            #ax.set_yticklabels((energies[idx]+0.5).astype(int))
```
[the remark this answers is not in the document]
daniloefl: Done.

JamesWrigley (approved):
This will need a changelog entry, otherwise LGTM! But I'm quite unfamiliar with this stuff and it was a pretty superficial review. I also worry a bit about so much untested code, it'd be great to have even simple smoke tests at some point.

daniloefl:
Thank you, @JamesWrigley . I have fixed the conflicts and added comments in the `changelog.md`. I will also follow up on tests in a new PR after discussing it with @takluyver .

PR #319, opened 2025-04-14
Delete nonexistent arguments from a docstring

JamesWrigley:
I'm guessing these are from a previous version of the code.


philsmt:
LGTM.

Yes, the recipes used to take run numbers and source names to construct their own `KeyData` objects.

JamesWrigley:
There was some resolver funkyness in CI so I took the liberty of bumping some constraints in 0c7cc57.

PR #320, opened 2025-04-16
Allow Scan.plot_bin_by_steps() to work for 2D binned results

takluyver:
In #269, we fixed the `bin_by_steps()` method to only average over the trains dimension, so you can get e.g. average spectra. But I didn't think about `plot_bin_by_steps()` at the time. This supports plotting the same information, mostly using xarray's defaults:

image


JamesWrigley (approved):
LGTM!

takluyver:
Thanks James!

PR #326, opened 2025-04-25
Add fel_shutter/ppl_shutter arguments to PumpProbePulses

philsmt:
I was discussing with @bermudei how to deal with arbitrary shutter blocking pulses in a pulse pattern, as occuring at FXE as well as SQS.

This is a proof-of-principle how one could integrate such support directly into `PumpProbePulses` via optional `fel_shutter`/`ppl_shutter` arguments that expect an `extra_data.KeyData` object (either directly or via alias). With this, one could do:

```python
run = open_run(<some-FEL-only-run>)
ppp = PumpProbePulses(
    run, pulse_offset=0,
    fel_shutter=run.alias['fel-shutter'],
    ppl_shutter=run.alias['ppl-shutter'])
```

As an extra bonus, this can avoid the "cannot infer PPL pulse position from FEL pulses" for PPL-only runs as long as the FEL was blocked and is not without pulses, which is generally often the case for such runs.

FYI @takluyver

### DIFF
--- a/src/extra/components/pulses.py
+++ b/src/extra/components/pulses.py
@@ -1258,6 +1258,10 @@ class PumpProbePulses(XrayPulses, OpticalLaserPulses):
             extrapolated from past trains if missing, true by default.
             An exception is raised if disabled and the PPL anchoring
             method is missing the minimum number of required FEL pulses.
+        fel_shutter (KeyData, optional): Scalar key evaluated as a
+            shutter state blocking FEL pulses in a given train.
+        ppl_shutter (KeyData, optional): Scalar key evaluated as a
+            shutter state blocking PPL pulses in a given train.
     """
 
     # This class inherits from two classes which both have the same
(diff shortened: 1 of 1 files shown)


takluyver:
Thanks, this looks like a neat extension. Unfortunately I suspect the shutter state is not normally saved, so we'd have to try to push them to include it in the DAQ.

I'm unsure if this would work when people use fast shutter manipulation to do train-alternating pump-probe, as happened in one experiment; I think that was also with a different laser from normal. But it would still be useful to know when the PP laser is blocked entirely.

I think there's also >1 shutter which can block each laser, although there's probably a specific one in each case that's normally used for that purpose, so we can avoid having to deal with that.

philsmt:
> I'm unsure if this would work when people use fast shutter manipulation to do train-alternating pump-probe, as happened in one experiment; I think that was also with a different laser from normal. But it would still be useful to know when the PP laser is blocked entirely.

That will depend on the implementation, though something changing every train will likely include some form of chopper wheel. At least the one I'm familiar with (HED's PPU) does not simply have a boolean key for its stage per train.

Anything implemented via some form of PLC switch should work though.

> I think there's also >1 shutter which can block each laser, although there's probably a specific one in each case that's normally used for that purpose, so we can avoid having to deal with that.

We could turn this into a list...

Alternatively, either one creates a custom FXE-specific version of `PumpProbePulses` that knows about the details, or we have a component managing shutter state with a suitable interface like `KeyData`.

philsmt:
@bermudei @takluyver given the time passed, can you confirm there's a point in finishing this?

PR #327, opened 2025-04-28
Add support for passing a union of runs to Scantool

JamesWrigley:
This is for convenience when working with combined runs. The output looks like this:
image

The bulk of the changes are to support reading the scantool settings from CONTROL data instead of RUN data.

### DIFF
--- a/src/extra/components/scantool.py
+++ b/src/extra/components/scantool.py
@@ -1,4 +1,5 @@
 import ast
+import logging
 from warnings import warn
 
 from extra_data import SourceData
@@ -43,14 +44,34 @@ class Scantool:
             else:
                 raise RuntimeError(f"Found multiple possible scantools, please pass one explicitly with the `src` argument: {', '.join(possible_devices)}")
 
-        values = run.get_run_values(src)
+        self._source_name = src
+        self._source = run[src]
+
+        # If the run is a union then we have to use the CONTROL values rather
+        # than RUN values.
+        run_values = run.get_run_values(src) if run.is_single_run else { }
+        if not run.is_single_run:
+            logging.warning("The passed DataCollection represents multiple runs, "
+                            "but this component will only take the Scantool settings from the first train.")
+
+        def get_value(key, raise_on_missing=True, is_str=False):
+            if key in run_values:
+                return run_values[key]
+            elif key in self._source:
+                arr = self._source[key][0].ndarray().squeeze()
+                return arr.item().decode() if is_str else arr
+            elif not raise_on_missing:
+                return None
+
+            raise KeyError(f"Could not find key '{key}' in either the RUN or CONTROL section")
 
         def get_first_value(keys):
             for key in keys:
-                if key in values:
-                    return values[key]
+                x = get_value(key, raise_on_missing=False)
+                if x is not None:
+                    return x
 
-            raise KeyError(f"Could not find any of these RUN section keys: {', '.join(keys)}")
+            raise KeyError(f"Could not find any of these keys in the RUN or CONTROL section: {', '.join(keys)}")
 
         # These are a list of possible property names for different versions of
         # the scantool. So far we've only seen the names being different, the
@@ -60,10 +81,8 @@ class Scantool:
         active_motors_keys = ["deviceEnv.activeMotors.value", "activeMotors.value"]
 
         # Get scan metadata and list of motors
-        self._source_name = src
-        self._source = run[src]
         self._active = self.source["isMoving"].ndarray().any()
-        self._scan_type = values["scanEnv.scanType.value"]
+        self._scan_type = get_value("scanEnv.scanType.value", is_str=True)
         self._motors = [x.decode() for x in get_first_value(active_motors_keys) if len(x) > 0]
 
         # The acquisition time vector gives the length of each step, unless the
@@ -73,14 +92,14 @@ class Scantool:
         # - https://git.xfel.eu/karaboDevices/Karabacon/-/blob/bd22d4a69bf7a401856f49920789ef42fda14ad2/src/karabacon/enums.py#L67
         self._acquisition_time = get_first_value(acquisition_time_keys)
         if _isinstance_no_import(self._acquisition_time, "numpy", "ndarray"):
-            if "Continuous" in values["deviceEnv.acquisitionMode.value"]:
+            if "Continuous" in get_value("deviceEnv.acquisitionMode.value", is_str=True):
                 self._acquisition_time = self._acquisition_time[0]
 
         # The deviceEnv.activeMotors property stores the motor aliases,
         # but we can try to get the actual device names from the
         # actualConfiguration property.
         self._motor_devices = None
-        motors_line = [x for x in values["actualConfiguration.value"].split("---") if "Motors:" in x]
+        motors_line = [x for x in get_value("actualConfiguration.value", is_str=True).split("---") if "Motors:" in x]
         device_names_warning = "Couldn't extract the Karabo device names for the active motors."
         if len(motors_line) == 1:
             try:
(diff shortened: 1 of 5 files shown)


src/extra/components/scantool.py
```diff
@@ -60,10 +81,8 @@ def get_first_value(keys):
         active_motors_keys = ["deviceEnv.activeMotors.value", "activeMotors.value"]
 
         # Get scan metadata and list of motors
-        self._source_name = src
-        self._source = run[src]
         self._active = self.source["isMoving"].ndarray().any()
-        self._scan_type = values["scanEnv.scanType.value"]
+        self._scan_type = get_value("scanEnv.scanType.value", is_str=True)
```
philsmt, line 85: Would `KeyData.as_single_value()` not work here?

JamesWrigley: That doesn't seem to work for strings, I get this error:

```python
TypeError                                 Traceback (most recent call last)
Cell In[6], line 1
----> 1 run["FXE_DAQ_SCAN[/MDL/KARABACON](https://max-jhub.desy.de/MDL/KARABACON)", "scanEnv.scanType.value"].as_single_value()

File [~/src/EXtra-data/extra_data/keydata.py:338](https://max-jhub.desy.de/user/wrigleyj/lab/tree/notebooks/src/EXtra-data/extra_data/keydata.py#line=337), in KeyData.as_single_value(self, rtol, atol, reduce_by)
    336     value = reduce_by(data)
    337 elif isinstance(reduce_by, str) and hasattr(np, reduce_by):
--> 338     value = getattr(np, reduce_by)(data, axis=0)
    339 elif reduce_by == 'first':
    340     value = data[0]

File ~/.conda/envs/scratch/lib/python3.12/site-packages/numpy/lib/_function_base_impl.py:4001, in median(a, axis, out, overwrite_input, keepdims)
   3916 @array_function_dispatch(_median_dispatcher)
   3917 def median(a, axis=None, out=None, overwrite_input=False, keepdims=False):
   3918     """
   3919     Compute the median along the specified axis.
   3920 
   (...)   3999 
   4000     """
-> 4001     return _ureduce(a, func=_median, keepdims=keepdims, axis=axis, out=out,
   4002                     overwrite_input=overwrite_input)

File ~/.conda/envs/scratch/lib/python3.12/site-packages/numpy/lib/_function_base_impl.py:3894, in _ureduce(a, func, keepdims, **kwargs)
   3891             index_out = (0, ) * nd
   3892             kwargs['out'] = out[(Ellipsis, ) + index_out]
-> 3894 r = func(a, **kwargs)
   3896 if out is not None:
   3897     return out

File ~/.conda/envs/scratch/lib/python3.12/site-packages/numpy/lib/_function_base_impl.py:4053, in _median(a, axis, out, overwrite_input)
   4049 indexer = tuple(indexer)
   4051 # Use mean in both odd and even case to coerce data type,
   4052 # using out array if needed.
-> 4053 rout = mean(part[indexer], axis=axis, out=out)
   4054 if supports_nans and sz > 0:
   4055     # If nans are possible, warn and replace by nans like mean would.
   4056     rout = np.lib._utils_impl._median_nancheck(part, rout, axis)

File ~/.conda/envs/scratch/lib/python3.12/site-packages/numpy/_core/fromnumeric.py:3860, in mean(a, axis, dtype, out, keepdims, where)
   3857     else:
   3858         return mean(axis=axis, dtype=dtype, out=out, **kwargs)
-> 3860 return _methods._mean(a, axis=axis, dtype=dtype,
   3861                       out=out, **kwargs)

File ~[/](https://max-jhub.desy.de/).conda/envs/scratch/lib/python3.12[/](https://max-jhub.desy.de/)site-packages/numpy/_core/_methods.py:147, in _mean(a, axis, dtype, out, keepdims, where)
    145         ret = ret.dtype.type(ret / rcount)
    146 else:
--> 147     ret = ret / rcount
    149 return ret

TypeError: ufunc 'divide' not supported for the input types, and the inputs could not be safely coerced to any supported types according to the casting rule ''safe''
```

takluyver:
Are the data you're pooling together several similar scans, or one scan split over multiple runs? Or something else? Even with the warning, I think this could be misleading - e.g. if you stick two scans together, should the number of steps be the total of the steps in each scan?

> The bulk of the changes are to support reading the scantool settings from CONTROL data instead of RUN data.

This feels somewhat awkward. Do we want a more general mechanism to get these kinds of values - either from RUN or CONTROL - when multiple runs are stuck together? Maybe with a check that the retrieved values are consistent across the pooled data? Or maybe just using `.as_single_value()` and fixing it for strings? (I think it predates the DAQ storing strings in CONTROL, so we probably didn't consider that)

JamesWrigley:
> Are the data you're pooling together several similar scans, or one scan split over multiple runs? Or something else? Even with the warning, I think this could be misleading - e.g. if you stick two scans together, should the number of steps be the total of the steps in each scan?

I think Ivette's usecase was with several similar scans? (correct me if I'm wrong @bermudei)
In either case, I would expect the `Scan` component to not know or care about the fact that the data is coming from multiple runs. If it's several similar scans then ideally it would combine the trains where the motor is at the same position (total number of steps $\approx$ number of steps in one run), and if it's one scan across multiple runs then I would expect the number of steps to be the total of the steps in each run.

> This feels somewhat awkward. Do we want a more general mechanism to get these kinds of values - either from RUN or CONTROL - when multiple runs are stuck together? Maybe with a check that the retrieved values are consistent across the pooled data? Or maybe just using `.as_single_value()` and fixing it for strings? (I think it predates the DAQ storing strings in CONTROL, so we probably didn't consider that)

Maybe? I dunno if it's worth it since AFAIK this is the only component where we had to look at run values, but if you think it'd be better I can have a go at implementing it.

bermudei:
Well, in the example I did yesterday for users there were 2 similar scan runs.. See 
below

 But I think they also did slightly different ones. So, James comment above should work for both cases. 
The reason why they needed this is for statistics...They couldnt really see any signal in a single run. It was tough :(

philsmt:
As Thomas puts it, string values have only been in `CONTROL` for a few months now. I made a draft on how we might be able to fix this here: https://github.com/European-XFEL/EXtra-data/pull/623

Concerning this PR, I'm hesitant to simply use the first value without any further check. This seems like baking in dangerous assumptions into a component that could break in mysterious ways. I agree when used properly this does the right thing, but that's not the situation I'm worried about. I'm afraid we risk the automagic behaviour we want to enable with components.

I fully agree with James that EXtra-data fundamentally motivates to treat multi-runs the same way as a single one. Code-wise however I would prefer to move any decision we make on how to treat this case upstream to EXtra-data, as it will come up more in other components. I already had cases where `AdqRawChannel` broke for multi-run `DataCollection` for the same reason. The `.is_single_run()` mechanism back then was meant as a first solution to not make it break in exactly  the mysterious ways mentioned above.

Maybe it's time to properly handle this instead?

JamesWrigley:
That's fair, though some properties like the acquisition time are stored as arrays, so `as_single_value()` would also need to be able to handle those. Alternatively, we could save the 'bounds' of each DataCollection that was unioned? That way we could check the settings at the first train of each sub-DataCollection, which would avoid having to load the data for all trains from disk. The downside is that it's not as robust as `as_single_value()` :thinking:

philsmt:
I was thinking in the direction of loading a `RUN` value for each file and do conflict resolution then. That should be cheap enough, a typical sequence is ~50s of measurement time.

tmichela:
Could an alternative be that the Scantool component accepts a list of runs rather than a union, such that you can read the `RUN` values for each and validate that they do not diverge?
Maybe that's something we could even upstream to EXtra-data (expose RUN data which is indentical in each of the merged DataCollection / raise if we acccess one that diverge).

PR #328, opened 2025-05-09
~~Fix~~ Prevent building with Cython 3.1

takluyver:
Use libc.math in Cython code instead of numpy.math. From [Cython's release notes](https://cython.readthedocs.io/en/latest/src/changes.html#beta-1-2025-04-03):

> The numpy.math cimport module has been deprecated. Usages should be replaced by libc.math. (Github issue [#6743](https://github.com/cython/cython/issues/6743))

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -39,7 +39,8 @@ Fixed:
   property for newer Scantool versions (!303).
 - Fixed [Scan.plot_bin_by_steps()][extra.components.Scan.plot_bin_by_steps] to show 2D
   data (!320).
-- Fixed building with Cython 3.1 (!328).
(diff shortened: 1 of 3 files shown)


philsmt (approved):
Thanks, LGTM

takluyver:
I fixed the original issue, where it failed to build, but ran into a new one: the signatures of functions accepting fused types are no longer exposed to Python. I've filed a Cython issue about this: https://github.com/cython/cython/issues/6855

In the meantime, I'll restrict it to build with Cython < 3.1.

philsmt:
Uh oh... One could get away with using `inspect.signature` on those functions, but it improves the error reporting. Let's try with Cython 3.0 for the time being, thanks.

PR #330, opened 2025-05-19
Allow building with Cython 3.1.2

takluyver:
Follow up to #328. The issue with inspecting function parameters when building with Cython 3.1.0 has been fixed.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,5 +1,11 @@
 [build-system]
-requires = ["setuptools>=61", "wheel", "setuptools_scm[toml]==7.1.0", "numpy", "cython<3.1"]
+requires = [
+    "setuptools>=61",
+    "wheel",
+    "setuptools_scm[toml]==7.1.0",
+    "numpy",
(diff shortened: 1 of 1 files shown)


philsmt:
Nice, thanks for taking care of the upstream part so quickly.

takluyver:
It was the Cython devs really, all I had to do was report it. :slightly_smiling_face: 

It looks like the docs build is broken, though...

```
_griffe.exceptions.AliasResolutionError: Could not resolve alias extra.signal._ftd.dled pointing at _cython_3_1_1.dled (in /home/docs/checkouts/readthedocs.org/user_builds/extra/envs/330/lib/python3.11/site-packages/extra/signal/_ftd.cpython-311-x86_64-linux-gnu.so:None)
```

takluyver:
I think the docs failure relates to a second issue in Cython around functions using fused types. I've reported this: https://github.com/cython/cython/issues/6897

takluyver:
OK, Cython 3.1.2 seems to resolve the issues we had, so I'll merge this.

PR #332, opened 2025-06-04
Align pulse information to the ADQ trains before using them

philsmt:
While https://github.com/European-XFEL/EXtra/pull/310 fixed the issue of pulse information *lacking* trains, it did not bother when pulse information has *more* trains. In combination with https://github.com/European-XFEL/EXtra/pull/245, it caused `AdqRawChannel.pulse_data()` to apparently return more trains than `AdqRawChannel.train_data()` as it's diligently filling anything with pulse information with fill values.

After fixing the above, I noticed the coordinates naturally required alignment as well. This PR addresses all of this issues by doing this upfront now, and added a test failing on the previous version.

### DIFF
--- a/tests/test_components_adq.py
+++ b/tests/test_components_adq.py
@@ -229,6 +229,22 @@ def test_pulse_data(mock_sqs_remi_run):
         # Test whether all pulses after the trace are nan.
         assert np.isnan(max_by_pulse[21:50]).all()
 
+    # Use pulse information with not all trains available.
+    ch = AdqRawChannel(
+        mock_sqs_remi_run, '3B', digitizer='SQS_DIGITIZER_UTC2',
+        pulses=XrayPulses(mock_sqs_remi_run.select_trains(np.s_[:-10])))
+
+    with pytest.raises(ValueError):
+        ch.pulse_data()
+
+    # Use pulse information with more trains available.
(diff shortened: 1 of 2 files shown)


tests/test_components_adq.py
```diff
@@ -229,6 +229,14 @@ def test_pulse_data(mock_sqs_remi_run):
         # Test whether all pulses after the trace are nan.
         assert np.isnan(max_by_pulse[21:50]).all()
 
+    # Use pulse information with not all trains available.
+    ch = AdqRawChannel(
+        mock_sqs_remi_run, '3B', digitizer='SQS_DIGITIZER_UTC2',
+        pulses=XrayPulses(mock_sqs_remi_run.select_trains(np.s_[:-5])))
+
+    with pytest.raises(ValueError):
+        ch.pulse_data()
```
takluyver, line 238: What's the ValueError that gets raised in this scenario?

philsmt: Good point, I didn't test the opposite fixed by this MR. Added now.

takluyver:
LGTM :+1: 

I think the coverage drop is a quirk of how codecov compares - it's comparing against the commit you branched from, and coverage has dropped in master since then.

PR #333, opened 2025-06-05
Add support for DataArrays to imshow2

JamesWrigley:
The advantage is that we can now get proper labels/axis limits automatically:
```python
plt.figure()
imshow2(azint.squeeze())
```
image


takluyver:
LGTM

PR #334, opened 2025-06-05
Always use the passed axis in Scan.plot_bin_by_steps()

JamesWrigley:
Otherwise it might use a different axis:
```python
fig, (ax1, ax2) = plt.subplots(1, 2, figsize=(9, 5))
scan.plot_bin_by_steps(results.intensity.squeeze(), ax=ax1)

fig.tight_layout()
```
image

This is a bit annoying to work around so I'm gonna go ahead and merge it :face_with_peeking_eye:

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -41,6 +41,9 @@ Fixed:
   data (!320).
 - Restrict the version of Cython used to build while we figure out an issue with
   Cython 3.1 (!328).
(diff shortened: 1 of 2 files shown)


takluyver:
Sorry, I must have missed this when I changed that method. I think we need to make a similar change in the 2D case, I can do a PR for that.

PR #349, opened 2025-06-19
Add reorder_axes_to_shape function

takluyver:
As we discussed, a function to reorder array axes to match a specified shape. This is meant to reduce the recurring issues where the array axis orders are not as expected, especially online.

Expected usage:

```python
jf_data = reorder_axes_to_shape(jf_data, (16, 512, 1024))

# With a wildcard
jf_data = reorder_axes_to_shape(jf_data, (None, 512, 1024))
```

The name is up for debate. We have a `reorder_axes` helper function in offline calibration (taking from & to axis names) which could also be added here, so I've gone for a similar name.

The wildcard could also be -1, like in `numpy.reshape()`. I don't have a strong preference.

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -69,3 +69,6 @@ def test_reorder_axes_to_shape():
 
     with pytest.raises(ValueError):
         reorder_axes_to_shape(arr, (None, None, 1024))  # Only 1 None allowed
+
+    with pytest.raises(ValueError):
+        reorder_axes_to_shape(arr, (None, 256, 1024))  # Wildcard & wrong number
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -1,7 +1,10 @@
 import numpy as np
+import pytest
 import xarray as xr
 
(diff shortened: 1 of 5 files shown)


tests/test_utils.py
```diff
@@ -38,3 +39,29 @@ def test_fit_gaussian():
     data = gaussian(np.arange(100), *params, norm=False)
     popt = fit_gaussian(data, A_sign=-1)
     assert np.allclose(popt, params)
+
+
+def test_reorder_axes_to_shape():
+    arr = np.zeros((512, 1024, 16), dtype=np.float32)  # E.g. burst mode JUNGFRAU data
+    res = reorder_axes_to_shape(arr, (16, 512, 1024))
+    assert res.shape == (16, 512, 1024)
+    assert res.base is arr
```
philsmt, line 48: Would you want to test if the data is actually transposed, as opposed to reshaped?

philsmt (approved):
Seems useful as you put it. Fine by me, one could add an actual test for correct data. LGTM.

PR #350, opened 2025-06-24
Access to dimension labels on calibration constants

takluyver:
Basic methods to retrieve the dimension labels that we're starting to store with constants, such as JUNGFRAU dark constants. Single and multi-module constant objects get a `.dimensions()` method, and `MultiModuleConstant.xarray()` will also use the dimension names if present, falling back to the `dim_0` style names it currently uses.

The way we've currently set up `SingleConstant`, getting the data and the dimension names will open and close the file twice; I haven't tried to optimise this away for now. For a multi-module constant we only do this with one module, and then assume the others are all the same - we could extend this if we want to change the axis order in the future, so you can mix old & new constants.

Questions:

- Detector components have a `.dimensions` property, whereas this is a `.dimensions()` method so you can optionally pass in `caldb_root`. Do we want to name it something different, and/or have a property which only works with the default caldb root path?
- Currently it returns `None` if the constant doesn't have dimension labels. Should it raise an error? Should it be possible to pass in a fallback to the method or the class?
- Do we want to integrate support for reordering dimensions, e.g. `consts['Offset'].ndarray(dim_order=["module", "cell", "slow_scan", "fast_scan", "gain"])` ? Or leave this to be done separately after loading?

### DIFF
--- a/src/extra/calibration.py
+++ b/src/extra/calibration.py
@@ -571,9 +571,10 @@ class MultiModuleConstant(Mapping):
         ndarr = self.ndarray(caldb_root, parallel=parallel)
 
         # Dimension labels
-        dims = self.dimensions(caldb_root) or (
-                ["module"] + ["dim_%d" % i for i in range(ndarr.ndim - 1)]
-        )
+        try:
+            dims = self.dimension_names(caldb_root)
+        except NoDimensionLabelsError:
+            dims = ["module"] + ["dim_%d" % i for i in range(ndarr.ndim - 1)]
         coords = {"module": modules}
         name = self.calibration_name
 


takluyver:
> I would prefer a property...
>
> I like raising an exception,

To my mind, these two preferences slightly contradict each other - if something can throw an exception in obvious cases, I tend to think it shouldn't be a property. I can live with that if you still think it's the best combo, though.

> isn't a different caldb root rare enough that we can force someone to go through the hassle of changing the default?

Probably, yes. I was trying to ensure there's always a way to do it without modifying global state, but possibly I'm being too purist about that.

> Would we then want to somewhere define the values to expect?

You mean the set of string labels we use? Yes, this should be documented in any case, and as we add it to more characterisation notebooks we should try very hard to keep it consistent.

philsmt:
LGTM

PR #353, opened 2025-07-02
Add a stub for extra-proposal

JamesWrigley:
I'm leaving this undocumented for now so that we have some freedom to change the API after some more testing.


takluyver:
LGTM

PR #354, opened 2025-07-03
Ridgeplot utility function

takluyver:
I can't explain this better than a screenshot:

image

### When to use a ridge plot?

- vs. plotting several lines on one axis: makes the order much easier to see (e.g. delay scan steps), but precise comparison of y coordinates is harder.
- vs. plotting each line on its own axis: shows more lines on a limited space.
- vs. a heatmap: the y scale is more explicit than a colour scale (e.g. this peak is twice that one), and it leaves room for drawing extra features like error bars or vertical spans. But a heatmap can display more data.
  - In some contexts I think one type of plot may just be more readily grasped. The example above is per-pulse digitizer traces; it feels more intuitive to show these as lines, even though I don't think there's any technical reason you can't use a heatmap.

### Why not use something existing?

There are a few Python packages to do ridgeplots, but none of them seem like a great fit. In particular, because this technique is often used for comparing distributions, the default is generally to treat the input data as samples for a [KDE](https://en.wikipedia.org/wiki/Kernel_density_estimation) before plotting it, and also to fill the area under the curve. I don't think that's mostly what we want.

### Questions & limitations

- Biggest question: can the utility function cover enough use cases to be worth having?
- Additional parameters for e.g. x & y limits? If you need to modify these at present, you have to iterate through the axes.
  - It could even return a wrapper class where e.g. `.set_xlim()` works on all the axes?
- Additional parameters for visual style, like the plot colours or the presence of the grey zero lines.
- Since it uses multiple matplotlib axes, you can't pass an axes object in, which most of our plotting functions allow. Should we allow passing a figure object instead? Or is there a way to take 1 axes and split it up?
- Drawing transparent things on overlapping axes can be a bit ugly:

image

### DIFF
--- a/src/extra/utils/misc.py
+++ b/src/extra/utils/misc.py
@@ -196,7 +196,25 @@ def hyperslicer2(arr, *args, ax=None, lognorm=False, colorbar=True, **kwargs):
     return controls
 
 
-def ridgeline_plot(data, *, overlap=0.5, xlabel=None, ylabel=None, stack_label=None, stack_ticklabels=None):
+def ridgeplot(data, *, overlap=0.5, xlabel=None, ylabel=None, stack_label=None, stack_ticklabels=None):
+    """Make a ridgeline plot showing a sequence of similar lines
+
+    A ridgeline plot spreads out the different lines vertically to make their
+    order clear, but allowing them to overlap. It's an alternative to a heatmap,
+    especially if there are relatively few rows (around 5-20).
+
+    Args:
+        data (array_like): A 2D array, each row of which will be plotted as one
+            line, starting at the top of the plot. Pass an xarray DataArray to
+            use its labels by default.
+        overlap (float): Number from 0 (no overlap) to 1, the fraction of each
+            plot's area covered by the next plot.
+        xlabel (str): Label for the shared x axis.
+        ylabel (str): Label for the y axis (drawn on the bottom plot).
+        stack_label (str): Label for the stacking axis (shown on the right)
+        stack_ticklabels (array_like): Labels for each line (shown on the right
+            next to the zero line of each plot).
+    """
     import matplotlib.pyplot as plt
     import matplotlib.gridspec as grid_spec
     from matplotlib.transforms import blended_transform_factory
--- a/src/extra/utils/misc.py
+++ b/src/extra/utils/misc.py
@@ -194,3 +194,65 @@ def hyperslicer2(arr, *args, ax=None, lognorm=False, colorbar=True, **kwargs):
         fig.colorbar(ax.get_images()[-1], ax=ax)
 
     return controls
+
+
+def ridgeline_plot(data, *, overlap=0.5, xlabel=None, ylabel=None, stack_label=None, stack_ticklabels=None):
+    import matplotlib.pyplot as plt
+    import matplotlib.gridspec as grid_spec
+    from matplotlib.transforms import blended_transform_factory
+
+    if data.ndim != 2:
+        raise TypeError(f"Expected a 2D array (got {data.ndim}D)")
+
+    gs = grid_spec.GridSpec(len(data), 1, hspace=-overlap)
+    fig = plt.figure(figsize=(8, 6))
+
+    if _isinstance_no_import(data, "xarray", "DataArray"):
+        x_data = data.coords[data.dims[1]]
+        xlabel = xlabel or data.dims[1]
+        stack_label = stack_label or data.dims[0]
+        if stack_ticklabels is None and data.dims[0] in data.coords:
+            stack_ticklabels = data.coords[data.dims[0]].values
+    else:
+        x_data = np.arange(data.shape[1])
+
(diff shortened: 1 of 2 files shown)


fadybishara:
Looks really cool, nice work! One thing to consider is to label the left $y$ axis with something like
```py
f"{quantity} Relative to Baseline [{units}]"
```

With or without the optional parameters `quantity` and/or `units`, or something similar.

src/extra/utils/misc.py
```diff
@@ -194,3 +194,105 @@ def hyperslicer2(arr, *args, ax=None, lognorm=False, colorbar=True, **kwargs):
...
+        data = np.asarray(data)
+    if data.ndim != 2:
+        raise TypeError(f"Expected a 2D array (got {data.ndim}D)")
+
+    if fig is None:
+        fig = plt.figure(figsize=(8, 6), layout="constrained")
+
+    gs = grid_spec.GridSpec(len(data), 1, hspace=-overlap)
+
+    if _isinstance_no_import(data, "xarray", "DataArray"):
```
fadybishara, line 238: It's not necessary but why not also allow `pandas.DataFrame`? It would require a little more handling than `xarray.DataArray`, that's for sure (I tried it quickly) but it would/could probably be useful?

tests/test_utils.py
```diff
@@ -21,6 +21,15 @@ def test_hyperslicer2():
     images = np.random.rand(10, 100, 100)
     hyperslicer2(images)
 
+
+def test_ridgeplot():
```
fadybishara, line 25: I made an random array and put some `np.nan`s in there; it threw a `ValueError` (see below) because `np.min((np.nan, 0))` returns `np.nan`, same with `np.max`. One very easy fix is to replace `np.min/max` with `np.nanmin/nanmax`; see [here](https://numpy.org/doc/stable/reference/generated/numpy.nanmin.html) but this of course doesn't cover other things that can go wrong :sweat_smile: 
```python
ValueError                                Traceback (most recent call last)
Cell In[36], line 1
----> 1 ridgeplot(tmparr)

Cell In[5], line 77, in ridgeplot(data, fig, overlap, xlabel, ylabel, ylim, yline, stack_label, stack_ticklabels)
     74 ax.axhline(yline, color='0.7', linewidth=1., zorder=0)
     76 # Use the same scale on each axes
---> 77 ax.set_ylim(y_min, y_max)
     78 ax.set_xlim(*x_range)
     80 if i < len(data) - 1:
     81     # Clear stuff from all but the bottom axes

File [/gpfs/exfel/sw/software/mambaforge/22.11/envs/202501/lib/python3.11/site-packages/matplotlib/axes/_base.py:3898](https://max-jhub.desy.de/gpfs/exfel/sw/software/mambaforge/22.11/envs/202501/lib/python3.11/site-packages/matplotlib/axes/_base.py#line=3897), in _AxesBase.set_ylim(self, bottom, top, emit, auto, ymin, ymax)
   3896         raise TypeError("Cannot pass both 'top' and 'ymax'")
   3897     top = ymax
-> 3898 return self.yaxis._set_lim(bottom, top, emit=emit, auto=auto)

File [/gpfs/exfel/sw/software/mambaforge/22.11/envs/202501/lib/python3.11/site-packages/matplotlib/axis.py:1210](https://max-jhub.desy.de/gpfs/exfel/sw/software/mambaforge/22.11/envs/202501/lib/python3.11/site-packages/matplotlib/axis.py#line=1209), in Axis._set_lim(self, v0, v1, emit, auto)
   1207 name = self._get_axis_name()
   1209 self.axes._process_unit_info([(name, (v0, v1))], convert=False)
-> 1210 v0 = self.axes._validate_converted_limits(v0, self.convert_units)
   1211 v1 = self.axes._validate_converted_limits(v1, self.convert_units)
   1213 if v0 is None or v1 is None:
   1214     # Axes init calls set_xlim(0, 1) before get_xlim() can be called,
   1215     # so only grab the limits if we really need them.

File [/gpfs/exfel/sw/software/mambaforge/22.11/envs/202501/lib/python3.11/site-packages/matplotlib/axes/_base.py:3585](https://max-jhub.desy.de/gpfs/exfel/sw/software/mambaforge/22.11/envs/202501/lib/python3.11/site-packages/matplotlib/axes/_base.py#line=3584), in _AxesBase._validate_converted_limits(self, limit, convert)
   3582     converted_limit = converted_limit.squeeze()
   3583 if (isinstance(converted_limit, Real)
   3584         and not np.isfinite(converted_limit)):
-> 3585     raise ValueError("Axis limits cannot be NaN or Inf")
   3586 return converted_limit

ValueError: Axis limits cannot be NaN or Inf
```

takluyver: Thanks, good catch. In my experience NaNs come up a whole lot more often than inf values, so I think using nanmin/nanmax is probably good enough for now.

src/extra/utils/misc.py
fadybishara: That makes sense, I agree with your logic.

PR #360, opened 2025-07-10
Allow to use DataCollection object as point in time for CalibrationData

philsmt:
It's not uncommon to look for calibration data for a particular run and one may want to take the run's beginning for the `event_at` (and `pdu_snapshot_at`) parameter. This MR adds support for this type by using the first train's timestamp.

This will also be used by a future `CalibrationData.from_data()` call.

### DIFF
--- a/src/extra/calibration.py
+++ b/src/extra/calibration.py
@@ -15,6 +15,9 @@ import pasha as psh
 import requests
 from oauth2_xfel_client import Oauth2ClientBackend
 
+from .utils.misc import _isinstance_no_import
+
+
 __all__ = [
     "CalCatAPIError",
     "CalCatAPIClient",
@@ -83,6 +86,8 @@ class CalCatAPIClient:
             return dt.astimezone(timezone.utc).isoformat()
         elif isinstance(dt, date):
             return cls.format_time(datetime.combine(dt, time()))
+        elif _isinstance_no_import(dt, 'extra_data', 'DataCollection'):
+            return cls.format_time(dt[0].train_timestamps(pydatetime=True)[0])
         elif dt is None:
             return ""  # Not specified - for searches, this usually means now
         elif not isinstance(dt, str):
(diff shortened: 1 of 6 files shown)


philsmt:
@JamesWrigley Somehow I made your test break by lifting `mock_spb_aux_run` to format version `1.2` 🤔

src/extra/calibration.py
```diff
@@ -588,6 +593,11 @@ def from_condition(
 
         `condition` should be a conditions object for the relevant detector type,
         e.g. `DSSCConditions`.
+
+        `event_at` and `pdu_snapshot_at` should either be an ISO 8601
+        compatible string or a datetime-like object. It may also be a
+        DataCollection object from EXtra-data to use the beginning of the
+        run as point in time.
```
takluyver, line 600:
```suggestion
        run as a point in time.
```

tests/test_calibration.py
```diff
@@ -235,3 +237,8 @@ def test_AGIPD_CalibrationData_report():
     assert set(agipd_cd) == {"Offset", "Noise", "ThresholdsDark", "BadPixelsDark"}
     assert agipd_cd.aggregator_names == [f"AGIPD{n:02}" for n in range(16)]
     assert isinstance(agipd_cd["Offset", "AGIPD00"], SingleConstant)
+
+
+def test_format_time(mock_spb_aux_run):
+    assert CalCatAPIClient.format_time(mock_spb_aux_run).startswith(
+        CalCatAPIClient.format_time(datetime.now())[:8])
```
takluyver, line 244: Do we also want to check that the formatted timestamp includes a timezone? I think checking that it contains one of `+zZ` ought to cover all possibilities.

takluyver (commented):
Other than that, LGTM

philsmt:
@JamesWrigley So I had a look what happens here.

It turns out the test was always broken and simply worked because format version 0.5 is hardcoded to report proposal/run -1 no matter what.

With the version fixed, it looked up the filenames and found r1 for both. While this was easy to fix and allowed the `assert not run.is_single_run` to pass, it still failed further below as no `MultiRunError` was thrown. This happens because those two runs actually don't share any XGM - constructing an `XGM` component on the union simply uses `SPB_XTD9_XGM/DOOCS/MAIN` from `moock_spb_aux_run` which is not in `multi_xgm_run` and thus rightfully throws no exception.

How do you want to fix this? I think `SPB_XTD9_XGM/DOOCS/MAIN` is the wrong name anyway, I suspect it's from some very old data, or copied from EXtra-data's mock runs.

JamesWrigley:
I see :thinking: You can comment it out for now, I'll take a look next week.

tests/test_calibration.py
philsmt: Good point, I refactored the test to actually use proper `datetime` objects. I figured relying on `format_time()` to return some ISO format is not a worse assumption than comparing strings...

PR #361, opened 2025-07-11
Fix pulse alignment when retrieving train and pulse edges

philsmt:
Adding to the endless saga, I forgot to also account for the edge retrieving methods in https://github.com/European-XFEL/EXtra/pull/332. This MR essentially adds the same fixes to those as well including the test.


takluyver:
LGTM

philsmt:
Thanks!

PR #362, opened 2025-07-11
Remove per-row index level across components

philsmt:
This is an API breaking change to correct one of my really stupid ideas.

The first components using `PulsePattern` components to label data added another index level for every individual row, e.g. `hitIndex` for hits, `signalIndex` for signals etc. This was meant to have an easy way to look into the first hit of every pulse, e.g. At the same time, this vastly complicates indexing pulse-resolved data with each other as `pandas` trips over the additional index level. Without this level, it's straightforward to index pulse-resolved data both with a train-resolved index as well as pulse-resolved index. The original functionality of getting the first hit for each pulse can easily be achieved using `.groupby()` if needed.

I also stopped using this level in `Timepix3`, so this change now fixes this for `DelayLineDetector` and `AdqRawChannel`.


docs/changelog.md
```diff
@@ -22,6 +22,8 @@ Changed:
   instead of zeros as a fill value for runs with a varying number of pulses
   (!347).
 - [imshow2()][extra.utils.imshow2] will now add a colorbar automatically (!351).
+- [DelayLineDetector][extra.components.DelayLineDetector] now longer attaches an index level for each row in [hits()][extra.components.DelayLineDetector.hits] and [signals()][extra.components.DelayLineDetector.signals] (!362).
+- [AdqRawChannel][extra.components.AdqRawChannel] now longer attaches an index level for each row in [train_edges()][extra.components.AdqRawChannel.train_edges] and [pulse_edges()][extra.components.AdqRawChannel.pulse_edges] (!362).
```
takluyver, line 26:
```suggestion
- [DelayLineDetector][extra.components.DelayLineDetector] no longer attaches an index level for each row in [hits()][extra.components.DelayLineDetector.hits] and [signals()][extra.components.DelayLineDetector.signals] (!362).
- [AdqRawChannel][extra.components.AdqRawChannel] no longer attaches an index level for each row in [train_edges()][extra.components.AdqRawChannel.train_edges] and [pulse_edges()][extra.components.AdqRawChannel.pulse_edges] (!362).
```

takluyver (commented):
I guess we can assume that not too many people have started using this code and relying on the extra index so far, so if there's any breakage beyond our own code, it should be minimal. :crossed_fingers:

philsmt:
I want to be honest here, this code has quite definitely been used over the past run in most applicable experiments. It is being advertised by example notebooks. I have not seen any use of this index level however, which coupled with the fact it makes indexing to trains/pulses harder is the reason I want to get rid of it. I waited for the end of the cycle/LIMP, so the 2025/1 environment remains compatible in case anyone relies on it.

takluyver:
Sorry, I wasn't writing very clearly. I understand that this breaks compatibility, and I think it's an acceptable trade-off. :+1:

philsmt:
@takluyver Was that an LGTM? 🙃

philsmt:
I decided to defer merging this MR towards the end of the cycle, to prevent the small off chance to break someone's analysis in the middle of a cycle.

PR #363, opened 2025-07-11
feat: jupyter widget for spectrometer calibration

tmichela:
https://github.com/user-attachments/assets/13a12bb2-584f-4236-a1f2-8636c8f8ecf0


bermudei:
Went through the code and tried the widget in a Jupyter notebook using FXE Jungfrau data. I didn't see any issues and was working as expected. I actually tried "breaking it" but didn't manage. So this looks good to me.

PR #368, opened 2025-08-12
Added more tests for the Grating.

daniloefl:
There have not been any tests for the `Grating*Calibration` classes. This adds some. It still relies on tests using data through https://git.xfel.eu/dataAnalysis/extra-tests/, but smaller tests are done anyway.


tests/conftest.py
```diff
@@ -158,3 +159,47 @@ def mock_timepix_exceeded_buffer_run(mock_sqs_timepix_directory):
...
+    # make it 2D
+    data2d = np.stack([data]*100, axis=1)
+    # add dimension for pulses in Gotthard
+    data_gh = np.stack([np.zeros_like(data)]*20+[data]*100, axis=-2)
+
+    sources = [
+        Timeserver('SQS_RR_UTC/TSYS/TIMESERVER'),
+        MonoMdl('SA3_XTD10_MONO/MDL/PHOTON_ENERGY', energy_data=energy),
+        CameraWithData("SQS_DIAG3_BIU/CAM/CAM_6", data=data2d),
+        GotthardIIWithData("SQS_EXP_GH2-2/CORR/RECEIVER", data=data_gh),
```
philsmt, line 189: Can you estimate how much data this is? We should be a bit careful with (actual) mockdata to not grow out of proportion.

philsmt (commented):
Thanks! Looks good to me in general, just would to doublecheck the amount of mockdata we're writing here with GH2.

tests/conftest.py
daniloefl: I have reduced the number of pixels in one of the dimensions by a factor of 10 and with that the tests still pass, and the temporary run produced is only 22 MB. I doubt this is a large amount and surely the resources available would be able to cope with it.

PR #373, opened 2025-08-22
Add Calibration.from_data()

philsmt:
One of the original ideas behind the `CalibrationData` API is to integrate the determination of detector condition from data, i.e. the various properties of control devices and/or the data itself. Due to various changes over the years to hardware and/or software, this can unfortunately be somewhat messy and is nowadays encoded in calibration notebooks, amongst other places.

This is an attempt to bring this functionality to the `CalibrationData` API in EXtra. Ideally, it makes obtaining calibration data as simply as:

```python
run = open_run(...)
cd = CalibrationData.from_data(run)
```

As it happens, the new design lends itself to this quite nicely by separating the detector-specific and -agnostic parts as before across the `CalibrationData` class and `ConditionsBase` implementation:

- Each detector-specific `ConditionsBase` object gains a `.from_data()` class method taking an `extra_data.DataCollection` as well as the CalCat detector name. From that, it attempts to initialize itself by inferring the necessary fields from the data. For `AGIPDConditions` e.g., its signature is:

  ```python
  @classmethod
  def from_data(cls, data, detector_name, modules=None,
                fpga_comp=None, mpod=None, fpga_control=None, xtdf=None,
                client=None, **params):
  ```
  `fpga_comp`, `mpod`, `fpga_control`, `xtdf` are inferred from the detector entry in CalCat and what's in `data`, or can be overwritten manually. The same applies to the `modules` list, which allows taking only the given modules into account (e.g. when a module is powered down, but part of the data).

  `**params` can be any of the regular parameter conditions, and prevents it from being inferred from data. The object attempts to be as greedy as possible and determine it from any applicable source, e.g. acquisition rate can be taken from the AGIPD FPGA comp device or the difference of pulse IDs in the data stream. If the sources in `data` are insufficient to construct the condition, an exception is thrown. This can be prevented by passing the missing parameters manually.

  For diagnostics, the actual extraction of any given parameter from data is implemented in its own static method, but that is an optional pattern.

- The `CalibrationData` class itself gains a `.from_data()` class method taking an `extra_data.DataCollection` and **optionally** a detector name. From that, it attempts to infer the detector name if missing via EXtra-data components, its detector type via CalCat and this pick the corresponding `ConditionsBase` type. Also, it infers `begin_at` from the data and thus attempt to initialize a proper .`CalibrationData` object via `.from_condition`. Any additional keyword arguments are passed to the conditions object, e.g. to overwrite parameters or sources.

For now, I implemented the necessary code to infer the condition for AGIPD, LPD and Jungfrau based on what we use in `pycalibration`, including (most) of its idiosyncracies. For AGIPD, I isolated the exact `pycalibration` code and ran it against this new API for a sample run of every proposal an AGIPD correction ever ran on with full agreement on all of them.

You can find a minimal example showcasing several situations [here](https://max-jhub.desy.de/user-redirect/notebooks/GPFS/exfel/data/scratch/schmidtp/caldata-examples.ipynb).

### DIFF
--- a/src/extra/calibration.py
+++ b/src/extra/calibration.py
@@ -1508,18 +1508,16 @@ class AGIPDConditions(ConditionsBase):
     def bias_voltage_from_control(sd):
         # These device used to suffer from switching to excessive values
         # randomly, so cut off any unreasonable values.
-        values = sd['highVoltage.actual'].ndarray()
+        values = sd['highVoltage.target'].ndarray()
         values = values[values < 1000.0]
         return int(np.median(values))
 
     @staticmethod
     def bias_voltage_from_mpod(sd, modules=None):
         if modules is not None:
-            keys = [f'channels.U{modno}.measurementSenseVoltage'
-                    for modno in modules]
+            keys = [f'channels.U{modno}.voltage' for modno in modules]
         else:
-            keys = sd.select_keys(
-                'channels.U*.measurementSenseVoltage').keys(False)
+            keys = sd.select_keys('channels.U*.voltage').keys(False)
 
         for key in keys:
             if (val := sd[key].as_single_value(atol=1, rtol=1)) > 0:
--- a/src/extra/calibration.py
+++ b/src/extra/calibration.py
@@ -1369,7 +1369,7 @@ class AGIPDConditions(ConditionsBase):
     acquisition_rate: float
     gain_setting: Optional[int]
     gain_mode: Optional[int]
-    source_energy: float
+    source_energy: float = 9.2
     integration_time: int = 12
     pixels_x: int = 512
     pixels_y: int = 128
--- a/src/extra/calibration.py
+++ b/src/extra/calibration.py
@@ -2,8 +2,9 @@
 import ast
 import json
 import re
-from collections.abc import Mapping
-from dataclasses import dataclass, field, fields, replace
+from collections.abc import Mapping, Iterable
+from dataclasses import (
+    dataclass, field, fields, replace, is_dataclass, MISSING)
 from datetime import date, datetime, time, timedelta, timezone
 from enum import IntFlag
 from fnmatch import fnmatch
@@ -21,6 +22,8 @@ import pasha as psh
 import requests
 from extra_data.read_machinery import find_proposal
 from oauth2_xfel_client import Oauth2ClientBackend
+
+from extra_data import PropertyNameError
 
 from .utils.misc import _isinstance_no_import
 
@@ -44,6 +47,14 @@ __all__ = [
 CALCAT_PROXY_URL = "http://exflcalproxy.desy.de:8080/"
 
 
+def any_is_none(*values):
+    for value in values:
+        if value is None:
+            return True
+
+    return False
+
+
 class ModuleNameError(KeyError):
     def __init__(self, name):
         self.name = name
@@ -52,9 +63,38 @@ class ModuleNameError(KeyError):
         return f"No module named {self.name!r}"
 
 
-
 class NoDimensionLabelsError(ValueError):
     pass
+
+
+class AutoConditionsError(ValueError):
+    """Used when detector conditions could not be inferred from data."""
+
+    def __init__(self, missing, sources):
+        self.missing = missing
+        self.sources = sources
+
+    def __str__(self):
+        msg = 'required parameters could not be inferred from data: ' + \
+            ', '.join(sorted(self.missing))
+
+        if self.sources:
+            msg += '\n\navailable sources to infer parameters:\n'
+
+            for name, value in self.sources.items():
+                if isinstance(value, Iterable) and not isinstance(value, str):
+                    if len(value) > 1:
+                        value = f'{sorted(value)[0]}, ...[{len(value) - 1} more]'
+
+                    elif len(value) == 1:
+                        value = str(next(iter(value)))
+
+                    elif not value:
+                        value = None
+
+                msg += f'- {name}: {value}\n'
+
+        return msg
 
 
 class CalCatAPIError(requests.HTTPError):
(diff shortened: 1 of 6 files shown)


src/extra/calibration.py
```diff
@@ -35,6 +39,14 @@
 CALCAT_PROXY_URL = "http://exflcalproxy.desy.de:8080/"
 
 
+def any_is_none(*values):
```
takluyver, line 42: Not remotely important to the overall PR, but this can be one line with the builtin [`any()`](https://docs.python.org/3/library/functions.html#any) function.

src/extra/calibration.py
```diff
@@ -199,6 +242,11 @@ def pdu_by_name(self, name):
         return self._get_by_name(
             "physical_detector_units", name, name_key="physical_name")
 
+    def pdus_by_detector(self, det_id, pdu_snapshot_at):
+        return self.get(f'physical_detector_units/get_all_by_detector',
+                        {'detector_id': det_id,
+                         'event_at': self.format_time(pdu_snapshot_at)})
```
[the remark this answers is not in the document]
philsmt: This discussion is now moot with the `DetectorData` API.

src/extra/calibration.py
```diff
@@ -1236,6 +1345,165 @@ def make_dict(self, parameters):
...
+    @classmethod
+    def from_data(cls, data, detector,
+                  fpga_comp=None, mpod=None, fpga_control=None, xtdf=None,
+                  client=None, **params):
+        # Uncaught exceptions that may be thrown in here:
+        # - NoDataError
+
+        if any_is_none(fpga_comp, mpod, fpga_control, xtdf):
+            if not isinstance(detector, DetectorData):
+                detector = DetectorData.from_identifier(detector)
```
turkot:
```suggestion
                pdu_snapshot_at = data.train_timestamps(pydatetime=True)[0]
                detector = DetectorData.from_identifier(
                    detector, pdu_snapshot_at=pdu_snapshot_at)
```
By default this would load current detector configuration, but I assume we want a snapshot from when the data was taken.

philsmt:
Rebased on top of the main branch after merging `DetectorData`, now to add some tests.

philsmt:
I've started adding tests replicating the ones done in `pycalibration` for operating condition detection from data. This has presented complications:

- The mock data framework always uses the current time as train timestamps, which ends up in the CalCat query and thus breaks the cassettes.
- All relevant parts of the mock data code is in EXtra-data, which makes adjusting it to cover more test cases difficult. 
- EXtra-data itself does not actually use that mock data for its tests

We have encountered the latter issue already in `pycalibration`. The proper solution to me would be to move all that mock data code to EXtra, including the mock devices only used in those. I would prefer to not grow this MR even more however, so I'd like to leave it with these tests for now and then move stuff over.

philsmt:
Rebased on top of `master` one more time

@takluyver I sneaked in a [minor fix](https://github.com/European-XFEL/EXtra/pull/373/changes/f2711bcd41add12c63240005afc190c6b6e9301d) to properly mark a regexp as raw string, `pytest` tipped me off.

src/extra/calibration.py
```diff
@@ -15,13 +16,16 @@
 from urllib.parse import urljoin
 from warnings import warn
 
+import numpy as np
```
turkot:
```suggestion
```
Numpy already imported on line 21.

src/extra/calibration.py
```diff
@@ -1304,6 +1413,171 @@ def make_dict(self, parameters):
...
+    def from_data(cls, data, detector,
+                  fpga_comp=None, mpod=None, fpga_control=None, xtdf=None,
+                  client=None, **params):
+        # Uncaught exceptions that may be thrown in here:
+        # - NoDataError
+
+        if any_is_none(fpga_comp, mpod, fpga_control, xtdf):
+            if not isinstance(detector, DetectorData):
+                detector = DetectorData.from_identifier(
+                    detector, pdu_snapshot_at=data)
```
turkot:
```suggestion
                detector = DetectorData.from_identifier(
                    detector, pdu_snapshot_at=data, client=client)
```

src/extra/calibration.py
```diff
@@ -1252,6 +1341,26 @@ def make_dict(self, parameters) -> dict:
...
+
+    @staticmethod
+    def _purge_missing_sources(data, *sources):
+        result = []
+
+        for src in sources:
+            if isinstance(src, str):
+                result.append(src if src in data.all_sources else None)
+            else:
+                result.append([s for s in src if s in data.all_sources])
```
turkot, line 1359: Shouldn't it append `None` in case `s` is not in `data.all_sources`?

src/extra/calibration.py
```diff
@@ -1349,6 +1623,71 @@ def make_dict(self, parameters):
...
         return cond
 
+    @classmethod
+    def from_data(cls, data, detector, fem_comp=None, xtdf=None,
+                  validate_memcell_order=False, use_memcell_order='auto',
+                  client=None, **params):
+        if any_is_none(fem_comp, xtdf):
+            if not isinstance(detector, DetectorData):
+                detector = DetectorData.from_identifier(
+                    detector, pdu_snapshot_at=data)
```
turkot:
```suggestion
                detector = DetectorData.from_identifier(
                    detector, pdu_snapshot_at=data, client=client)
```

src/extra/calibration.py
```diff
@@ -1486,6 +1852,95 @@ def make_dict(self, parameters):
 
         return cond
 
+    @classmethod
+    def from_data(cls, data, detector, control=None, client=None, **params):
+        if control is None:
+            detector = DetectorData.from_identifier(detector, client=client,
+                                                    pdu_snapshot_at=data)
```
turkot, line 1858: Contrary to `AGIPDConditions` and `LPDConditions`, here `client` is passed 🙂

turkot:
I've tested on:
- AGIPD p010627 r0241
- Jungfrau p009056 r0076
- LPD p007944 r0346, p003409 r0151

Works fine :+1: 

Does not work for AGIPD p700000  r0030 (p2120 r0097) from 2018, which is fine :slightly_smiling_face:

src/extra/calibration.py
turkot: Never mind, in the worst case it will append an empty list.

src/extra/calibration.py
philsmt: Oops, I think I caused this while rebasing. Thanks!

src/extra/calibration.py
philsmt: Well spotted!

src/extra/calibration.py
philsmt: 🙈

philsmt:
Thanks for spotting those remaining blunders!

turkot:
LGTM! Can we have it in our environment? :pleading_face:

philsmt:
It will be the day after it is merged.

PR #379, opened 2025-09-15
Add method to group xarray by scan steps

takluyver:
This is similar to `Scan.split_by_steps`, but:

1. It produces an xarray GroupBy object, rather than just a sequence of smaller DataArrays. This has shortcuts for e.g. getting the mean of each group as a new xarray.
2. It works if the trainId coordinate is one level in a MultiIndex. I don't much like the way I've done this, but I haven't found a neater way - ideas welcome.

This is based on a function I wrote in fxeutils, but generalised a bit.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -25,6 +25,8 @@ Added:
   supports `DataCollection` objects to reference a point in time
 - Pre-built packages will be available on PyPI for Python 3.10 - 3.13 from the
   next release, rather than only Python 3.10 (!377).
+- [Scan.group_xarray] method to make an xarray GroupBy object based on scan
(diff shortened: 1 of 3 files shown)


philsmt (approved):
Thanks, that's an awesome idea. I tried around myself a bit and did not find a better solution that will work across all edge cases of train alignment etc. So LGTM for the implementation.

I would definitely want to extend this to `pandas` types, and unfortunately the APIs for their respective `.groupby()` seems to differ for complex cases 😒 

With the above I might want to have a method `.group()` that just magically works across types, but it's fine if you want to stick with including the `_xarray` label for now.

PR #385, opened 2025-09-25
Allow passing DataArrays to fit_gaussian() and add `nans_on_failure`

JamesWrigley:
I usually end up doing these two things anyway. Also sneaked in a wee fix for mkdocs in 6198f7c.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -27,6 +27,10 @@ Added:
   next release, rather than only Python 3.10 (!377).
 - [Scan.group_data] method to make an xarray or pandas GroupBy object based on
   scan steps (!379).
+- [fit_gaussian()][extra.utils.fit_gaussian] has a new `nans_on_failure`
+  argument to allow returning NaN arrays when fitting fails (!385).
+- [fit_gaussian()][extra.utils.fit_gaussian] now supports
+  [DataArray][xarray.DataArray] inputs (!385).
 
 Changed:
 
--- a/src/extra/utils/functions.py
+++ b/src/extra/utils/functions.py
@@ -1,5 +1,7 @@
 import numpy as np
 
+from .misc import _isinstance_no_import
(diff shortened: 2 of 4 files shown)


JamesWrigley:
(if there's no objections I'll merge this by the end of the week)

src/extra/utils/functions.py
```diff
@@ -45,20 +48,33 @@ def fit_gaussian(ydata, xdata=None, p0=None, norm=False, A_sign=0, **kwargs):
...
     """
     if xdata is None:
         xdata = np.arange(len(ydata))
 
+    if _isinstance_no_import(ydata, "xarray", "DataArray"):
+        ydata = ydata.values
+    if _isinstance_no_import(xdata, "xarray", "DataArray"):
+        xdata = xdata.values
+
+    failure_value = np.full(len(p0), np.nan) if nans_on_failure else None
```
takluyver, line 76: If p0 is passed, it has to have length 4, right? Perhaps we want to check it and throw an error if not, but otherwise I don't think passing a different length list should mean a change in the return value.

```suggestion
    failure_value = np.full(4, np.nan) if nans_on_failure else None
```

docs/changelog.md
```diff
@@ -27,6 +27,9 @@ Added:
   next release, rather than only Python 3.10 (!377).
 - [Scan.group_data] method to make an xarray or pandas GroupBy object based on
   scan steps (!379).
+- [fit_gaussian()][extra.utils.fit_gaussian] has a new `nans_on_failure`
+  argument to allow returning NaN arrays when fitting fails, and now also
+  supports [DataArray][xarray.DataArray] inputs (!385).
```
takluyver, line 32: I'd maybe put these as separate bullet points, since they seem to be orthogonal changes that you just happened to make together.

PR #417, opened 2025-11-18
Add pulse selection machinery and manual pulse patterns

philsmt:
The existing `PulsePattern`-based components can expose pulse patterns as described by data, but so far it was not possible to modify those patterns. Conceptually it seemed rather convoluted to track modifications from the data state. Independent of that, I meant to have some `ManualPulses` component for some time that exposes the same interface but allows to make up your own patterns. Turns out those two concepts actually are made for each other!

This PR adds said `ManualPulses` type that simply holds any desired pulse pattern internally without any data reference, and then implements `select_pulses()`, `deselect_pulses()` and `union()` on top of that - applying any such modification to a `PulsePattern` component will always return a `ManualPulses` object indicating it is no longer represents a pattern found in data. 

Furthermore, I tried to improve the documentation on these components with some introduction and examples.

### DIFF
--- a/src/extra/components/pulses.py
+++ b/src/extra/components/pulses.py
@@ -129,16 +129,21 @@ class PulsePattern:
     """Abstract interface to pulse patterns.
 
     This class should not be instantiated directly, but one of its
-    implementationsd `XrayPulses` or `OpticalLaserPulses`. It provides
-    the shared interface to access any pulse pattern.
-
-    Instances of this class are assumed to be immutable, which is
-    exploited in various methods for caching purposes. An implementation
-    should not allow altering its internal object state in a way that
-    causes return values of public APIs to change.
-
-    Requires to implement _get_pulse_ids().
+    implementations such as `XrayPulses` or `OpticalLaserPulses`. It
+    provides the shared interface accessible from all oher pulse pattern
+    implementations, however.
     """
+
+    # Instances of this class are assumed to be immutable, which is
+    # exploited in various methods for caching purposes. An implementation
+    # should not allow altering its internal object state in a way that
+    # causes return values of public APIs to change.
+
+    # The minimum requirement for a custom pulse pattern type is to
+    # implement the abstract method `_get_pulse_ids()`.
+    # If the underlying pattern may also describe pulse-less trains,
+    # `_get_train_ids()` should be overwritten as well. Its default
+    # implementation will return only the train IDs with pulses.
 
     # Number of elements in bunch pattern table according to XFEL Timing
     # System Specification, Version 2.2 (2013). The original table may
@@ -964,12 +969,14 @@ class TimeserverPulses(PulsePattern):
     """Abstract interface to timeserver-based based pulse patterns.
 
     This class should not be instantiated directly, but one of its
-    implementations `XrayPulses` or `OpticalLaserPulses`. It provides
-    the shared interface to access pulse patterns encoded in the bunch
-    pattern table.
-
-    Requires _mask_table() and _get_ppdecoder_nodes() to be implemented.
+    implementations such as [`XrayPulses`][extra.components.XrayPulses]
+    or [`OpticalLaserPulses`][extra.components.OpticalLaserPulses]. It
+    provides the shared interface to access pulse patterns encoded in
+    the timeserver or pulse pattern decoder data.
     """
+
+    # An implementation os required to implement _mask_table() and
+    # _get_ppdecoder_nodes().
 
     # Timeserver class ID and regular expressions.
     _timeserver_class = 'TimeServer'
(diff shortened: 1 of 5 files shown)


src/extra/components/pulses.py
```diff
@@ -1508,6 +1763,96 @@ def pulse_mask(self, labelled=True, field=None):
...
+                Train IDs or data to take train IDs from.
+            pulse_ids (numpy.typing.ArrayLike): Pulse IDs of a single
+                train to be repeated in every train.
+
+        Returns:
+            pulses (ManualPulses): Manual pulse pattern with specified
+                pulse IDs in every train.
+        """
+
+        if isinstance(train_ids, TrainData):
```
fadybishara: If `isinstance(train_ids, TrainData)` is False, the `train_ids` can have any type*. It seems to me that this is not the desired behavior, is it?

\* If we run a type checker, which AFAIK we don't, then any type is clearly any type that matches `np.typing.ArrayLike` but still, this means a list of strings would be okay.

src/extra/components/pulses.py
```diff
@@ -10,38 +10,140 @@
...
+                # here, modify the slice accordingly.
+                pulse_sel = slice(
+                    pulse_sel.start, pulse_sel.stop - 1, pulse_sel.step)
+
+            return pulse_ids.loc[train_sel, pulse_sel]
+
+        def slice_pulses(pulse_ids):
+            return pulse_ids.iloc[pulse_sel]
+
+    elif isinstance(pulse_sel, (list, np.ndarray)):
```
fadybishara: Here it explicitly allows a List for `pulse_sel` so comment above stands I think...

src/extra/components/pulses.py
philsmt: Hmmm, so we generally don't check such values for `ArrayLike` everywhere unless we distinguish between different collection types, but allow it to fail wherever it does. There's a point one could do this more stringently, but are *ArrayLike* checks nowadays flexible enough to catch any such type in the wild?

Given Python's history of duck typing, I would be cautious with hard checks unless it's strictly required - here it's essentially just `len()` and iterability.

src/extra/components/pulses.py
```diff
@@ -1508,6 +1763,96 @@ def pulse_mask(self, labelled=True, field=None):
...
+        if repetition_rate is not None:
+            pulse_period = int(round(1.3e9 / 288 / repetition_rate , 0))
+        elif pulse_period is None:
+            raise ValueError('must specify either repetition_rate or '
+                             'pulse_period')
+
+        pulse_ids = np.arange(pulses_per_train) * pulse_period \
+            + pulse_id_offset
+
+        return cls.from_constant_pulses(train_ids, pulse_ids)
```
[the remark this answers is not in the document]
philsmt: You mean to check whether `len()` can be called on it? The docstring indicates *ArrayLike*.

src/extra/components/pulses.py
```diff
@@ -10,38 +10,140 @@
...
+
+    if train_sel is not None:
+        from extra_data.read_machinery import select_train_ids
+        train_sel = select_train_ids(
+            # Select on actual train IDs with pulses.
+            list(pulse_ids.index.to_frame()['trainId'].unique()),
+            train_sel)
+
+    else:
+        train_sel = np.s_[:]
```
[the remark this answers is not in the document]
philsmt: Personal flavour, I prefer `np.s_[]` over `slice()` for slice literals as it maintains the slicing syntax.

src/extra/components/pulses.py
```diff
@@ -10,38 +10,140 @@
...
+
+    if isinstance(pulse_sel, by_index):
+        # Convert by_index to a regular slice.
+        pulse_sel = pulse_sel.value
+
+    if isinstance(pulse_sel, by_id):
```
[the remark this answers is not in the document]
philsmt: `by_id` meant to use as an indexer, but calling it also works:

```ipython
In [2]: isinstance(x := by_id[[100, 101]], by_id), x.value
Out[2]: (True, [100, 101])

In [3]: isinstance(x := by_id([100, 101]), by_id), x.value
Out[3]: (True, [100, 101])
```

src/extra/components/pulses.py
philsmt: You mean it could be generalized more?

The idea is to support the same types that could be passed to EXtra-data, i.e. those resolved by `extra_data.read_machinery.select_train_ids`. There it explicitly checks for `slice`, `list`, `np.ndarray`. You're right the code could work with many more, but given it goes through a custom object like `by_id[]`, it seems acceptable to restrict it a narrower range of documented types.

But this is debatable.

src/extra/components/pulses.py
fadybishara: Okay, I was just worried that if the `isinstance` check fails, then the sky is the limit. But I think I miunderstood your rationale for this check, it's not a negative but, rather, a positive one, no? I mean if it evaluates to `True`, then we have more information to work with. Is this correct? If so, then it's fine.

src/extra/components/pulses.py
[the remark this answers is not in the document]
fadybishara: I think this comment is related the last one so I will mark it as resolved.

fadybishara:
LGTM! Feel free to merge :smiley:

philsmt:
Thanks for review!

PR #430, opened 2025-12-12
Add CalibrationData.reports_info() method

takluyver:
This is meant to give a condensed summary of where a group of constants have come from:

image

I'll open this as a draft to remind us that #408 and #373 are going first.


philsmt (approved):
Thanks, looks nice and useful. LGTM!

PR #452, opened 2026-02-05
Increase bunch pattern table length to 4096

philsmt:
This accomodates a change done do the underlying timing system in anticipation of longer train lengths enabled by GUN5.


JamesWrigley (approved):
LGTM!

PR #476, opened 2026-04-07
Improve robustness of the calibration with *very* poor quality data.

daniloefl:
The goal of this MR is to make the calibration algorithm in `CookieboxCalibration` more robust. In the past, the data used for the calibration were relatively clean, but in recent analyses with low pulse energy, this is not so.

The changes done here include:
- RANSAC is used instead of a simple ordinary least squares line fit, which is much more robust to outliers, and also derives a mask from the data itself. The user can still provide a mask, but it is updated with the result of RANSAC. This leads to a better fit performance even with a large amount of outliers.
- The Golden method is used for $t_0$ search instead of the Brent method. This search strategy is more robust to outliers as well.
- The option `count_threshold` has been added and if it is set to a negative value, photo-electron counting is performed to select good quality data instead of a cut on XGM. This leads to a significantly better quality of data cleaning.
- If the peak-finding stage fails, mark that peak to be masked. This adds an extra stage of checks for bad quality data.

Additionally, some old, unused code has been removed, namely the "per-shot" normalization. This is meaningless after the previous MR in `CookieboxCalibration` and it was forgotten to be removed there.

The response function estimate in `TofAnalogResponse` also has the option `count_threshold` to use photo-electron counting to obtain a more reliable result in very noisy environments.


src/extra/applications/cookiebox.py
```diff
@@ -114,23 +119,22 @@ def fit(peak_ids: np.ndarray, energies: np.ndarray, t0_bounds: Tuple[float, floa
...
-    else:
-        raise Exception('fit did not converge.')
+    c, e0, t0, mask = lin_fit(peak_ids, energies, t0)
+    if not res.success:
+        return np.nan, np.nan, np.nan, mask
+    return c, e0, t0, mask
 
 
-def calc_mean(itr: Tuple[int, int], scan: Scan, xgm_data: xr.DataArray, tof: Dict[int, AdqRawChannel], xgm_threshold: float, correction_fn=None) -> xr.DataArray:
+def calc_mean(itr: Tuple[int, int], scan: Scan, xgm_data: xr.DataArray, tof: Dict[int, AdqRawChannel], xgm_threshold: float, count_threshold: float=None, correction_fn=None) -> xr.DataArray:
```
JamesWrigley: What if this defaulted to 0 instead of `None`? Then we would only have to check if `count_thresholds >=0`.

src/extra/applications/cookiebox.py
```diff
@@ -140,29 +144,49 @@ def calc_mean(itr: Tuple[int, int], scan: Scan, xgm_data: xr.DataArray, tof: Dic
...
-    out_xgm = tof_xgm_data.mean('pulse')
+    # check if we are required to count peaks
+    if count_threshold is None or count_threshold >= 0:
+        # don't count peaks and just average the data
+        # the logic to clean the data is only to apply a cut on the XGM pulse energy
+
+        # this does train ID matching, because:
+        # - the given run may not have both XGM and eTOF for every train;
+        # - and we cannot control the creation of the run object, since we want to receive the ready-made XGM object
+        good_ids = sorted(list(set(train_ids).intersection(set(xgm_data.coords["trainId"].to_numpy()))))
```
JamesWrigley: Nitpicking, this is a slightly shorter way of getting the coordinates:
```suggestion
        good_ids = sorted(list(set(train_ids).intersection(set(xgm_data.trainId.data))))
```

src/extra/applications/cookiebox.py
```diff
@@ -140,29 +144,49 @@ def calc_mean(itr: Tuple[int, int], scan: Scan, xgm_data: xr.DataArray, tof: Dic
...
+        out_data = -tof_data.mean('pulse')
+        out_xgm = tof_xgm_data.mean('pulse')
+    else:
+        # option 2: count photon peaks
+        # in this case, ignore the XGM, as it is only used for cleaning the data
+        # and here we rely on the threshold for that
+        tof_data = tof[tof_id].select_trains(by_id[train_ids])
+        tof_data = tof_data.pulse_edges(pulse_dim='pulseIndex', threshold=count_threshold).reset_index()
+
+        bins = np.arange(0, 500+1)
```
JamesWrigley: Should the 500 be coming from an argument?

JamesWrigley (changes requested):

src/extra/applications/cookiebox.py
daniloefl: Well, actually the threshold is often negative and could be zero. I wanted to create a very clear distinction for that reason. I could also use `np.nan`, but I thought `None` was simpler.

src/extra/applications/cookiebox.py
daniloefl: Indeed that is cleaner. I just did that. The concern is that this number needs to be the same for all of the parallel calls of `calc_mean` (it defines the length of an axis, which later gets stacked). I have not seen this ever needing to be bigger than 500, so I would keep it like this, instead of adding yet another user parameter for something that is unlikely to ever change. But at least like this, I can document it.

What do you think? I could also provide it entirely as a user parameter, but in this case it just seems like overkill.

JamesWrigley: Ah I see :thinking: Maybe make it a global constant then?

JamesWrigley: Oh sorry I think I misunderstood what you meant by parallel calls, do you mean a multiprocessing thing?

I originally noticed it because of the `500 + 1` construction instead of `501`, which made me think the `500` was meant to be an argument. Is that just to make clear that there should 500 (plus an edge?) bins?

daniloefl: Yes, indeed: they are 500 bins plus edge. And this function is called in parallel, as you mentioned.

I think keeping it as an argument documents it better, and if I think later that it is good to change it (for a reason I don't yet see), it could be easier to extend it.

daniloefl: I have now added a parameter in the constructor to handle this case.

JamesWrigley (approved):
LGTM!

PR #477, opened 2026-04-09
Make Scan object from motor target positions

takluyver:
This is a request from FXE: rather than defining a scan just from motor actual positions, which can be noisy, it can be convenient to define it based on target positions, which are in neat steps, and exclude points where the actual position is too far from the target.

This would be used like:

```python
scan = Scan.from_motor_targets(run['FXE_XTD9_MONO/MDL/ACCM_PITCH'], tolerance=1e-4)
```

image

We could probably guess a default tolerance, e.g. 20% of the smallest step size, but for now I haven't tried to do this.

TBD: tests

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -32,6 +32,9 @@ Added:
   next release, rather than only Python 3.10 (!377).
 - [Scan.group_data()][extra.components.Scan.group_data] method to make an xarray
   or pandas GroupBy object based on scan steps (!379).
+- [Scan.from_motor_targets()][extra.components.Scan.from_motor_targets] method
+  to make a scan object based on the target positions of the motor, excluding
+  points where the actual position is too far from the target.
 - [fit_gaussian()][extra.utils.fit_gaussian] has a new `nans_on_failure`
   argument to allow returning NaN arrays when fitting fails (!385).
 - [fit_gaussian()][extra.utils.fit_gaussian] now supports
--- a/src/extra/components/scan.py
+++ b/src/extra/components/scan.py
@@ -31,7 +31,7 @@ class Scan:
     ```
     """
 
-    def __init__(self, motor, name=None, resolution=None, min_trains=None, intra_step_filtering=1):
+    def __init__(self, motor, name=None, resolution=None, min_trains=None, intra_step_filtering=1, *, _steps=None):
         """The class tries to detect the best parameters automatically, but it will
         still fail in some situations. If that happens either the `resolution`
         or `min_trains` parameters (unlikely to be both) will need to be set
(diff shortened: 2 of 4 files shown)


philsmt:
Oh, this is a really nice idea! Would you consider a more generic approach by allowing another `motor`-like object to be passed to `Scan`'s initializer for this purpose? It doesn't seem like your implementation doesn't make any assumptions about literal motors.

The rationale is that I had a quite similar problem recently when looking at delay scans - you want to use BAM/LAM/PAM-corrected data as position, but this is too noisy to do the actual step separation.

If you'd like to convenience of passing `SourceData` instead of keys for actual motors, maybe we can generally make a static method for this purpose.

takluyver:
Sure, I'm open to more flexibility. What do you have in mind? You pass in separate but aligned data for identifying steps and filtering tolerance? Or does it need 3 categories, to also control how you find the position of each step? E.g.

```python
# Delay scan
Scan(
    step_data=delay_line_target_delay,
    position_data=pam_measured_delay,
    tolerance_data=pam_measured_delay,
    tolerance=1e-3,  #Difference between position data & tolerance data
)

# Motor scan
Scan(
    step_data=motor['targetPosition'],
    position_data=motor['targetPosition'],
    tolerance_data=motor['actualPosition'],
    tolerance=1e-4,
)
```

(That's just to give a rough idea - if we actually implement something like this, we might want it to fit in better with the existing API)

I suspect that scanning a motor, or a motor-like device (`FXE_XTD9_MONO/MDL/ACCM_PITCH` isn't really a motor, but it has targetPosition & actualPosition) is common enough that it's probably nice to have a shortcut for that.

philsmt:
Hmmm, I was thinking that the delay stage is identical to your target, and the total measured delay value akin to your actual. Then one can use the scalar tolerance in the same way to ensure they don't stray too far. I'm not sure one would need separate tolerance data 🤔 

Ignore my static helper method point, I just realized `Scan` already supports that via passing a `SourceData` object. Essentially my point is to move your functionality into the initializer and add `target` as a new optional argument. If not given, `target = motor`.

We could also rename `motor` while we're at it to `actual`, and potentially add a keyword-only `motor` argument for backwards compatibility in case someone used it in this way.

takluyver:
Something like this?

I've reused the resolution parameter as equivalent to tolerance, rather than adding another parameter. When you've got target positions, which should be less noisy, I've arbitrarily picked a default of 1/5 the median step size. :shrug: 

I've left a static method separate as an opt-in to using this. If we start automatically using `targetPosition` from a SourceData object, that will change the behaviour of existing code using `Scan`.

philsmt:
Yes, thanks. That should work nicely for general use cases.

I don't have any data to comment on re-using `resolution` over a dedicated argument, so let's go with it for now.

PR #479, opened 2026-04-10
Add source names to DetectorData

philsmt:
Adds support for source names to new `DetectorData` API based on the work already done in https://github.com/European-XFEL/EXtra/pull/146

Given we just merged the API yesterday, I'll omit a changelog entry unless this lingers for a long time.

### DIFF
--- a/src/extra/calibration.py
+++ b/src/extra/calibration.py
@@ -1,3 +1,5 @@
+
+import ast
 import json
 import re
 from collections.abc import Mapping
@@ -8,6 +10,7 @@ from fnmatch import fnmatch
 from functools import lru_cache
 from operator import index
 from pathlib import Path
+from string import Formatter
 from typing import Dict, List, Optional, Union
 from urllib.parse import urljoin
 from warnings import warn
@@ -1368,6 +1371,21 @@ class ShimadzuHPVX2Conditions(ConditionsBase):
     }
 
 
+class SourceExprChecker(ast.NodeVisitor):
+    def visit_Call(self, node):
+        raise ValueError("Function calls not allowed in source name patterns")
+
+
+class SourceNameFormatter(Formatter):
+    """String formatter that evaluates simple operations like {modno + 2}"""
+
+    def get_field(self, field_name, args, kwargs):
+        node = ast.parse(field_name, "<source pattern>", "eval")
+        SourceExprChecker().visit(node)
+        obj = eval(compile(node, "<source pattern>", "eval"), kwargs)
+        return obj, 0
+
+
 @dataclass
 class DetectorModule:
     """Detector module.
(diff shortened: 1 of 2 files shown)


src/extra/calibration.py
```diff
@@ -1402,12 +1420,17 @@ class DetectorModule:
     module_number: int | None
     detector_type: str
     legacy_uuid: int | None  # Deprecated, do not use
+    source_name: str
```
takluyver: We're allowing for it being None in the code, so the type hint should reflect that.

```suggestion
    source_name: str | None
```

src/extra/calibration.py
```diff
@@ -1402,12 +1420,17 @@ class DetectorModule:
...
 
     def __post_init__(self):
         if self.module_number is None:
             # Try to fill in module number if missing.
             self.module_number = int(re.findall(r"\d+", self.aggregator)[-1])
 
+        if self.source_name is not None:
+            self.source_name = SourceNameFormatter().format(
+                self.source_name, modno=self.module_number)
```
takluyver, line 1432: I think it would be cleaner to expand the template before constructing a `DetectorModule`, rather than setting the value to the template initially and then changing it from `__post_init__`.

tests/test_calibration.py
```diff
@@ -399,12 +400,15 @@ def test_DetectorData_from_identifier():
     assert list(agipd)[0] == next(iter(agipd.keys())) == 'AGIPD00'
     assert isinstance(next(iter(agipd.values())), DetectorModule)
     assert agipd[0] == agipd['AGIPD00']
+    assert all([name == f'SPB_DET_AGIPD1M-1/DET/{i}CH0:xtdf' for i, name
+                in enumerate(agipd.source_names)])
```
takluyver:
```suggestion
    assert agipd.source_names == [
        f'SPB_DET_AGIPD1M-1/DET/{i}CH0:xtdf' for i in range(len(agipd.source_names)
    ]
```

If it fails, pytest gives more useful information from comparing two lists than from doing `all()` on a list of comparisons (I just tested, the `assert all(...)` shows you which of the first 5 failed, but not the values compared). Not that I think this will ever fail, but...

src/extra/calibration.py
philsmt: I agree it's not super elegant, but that would make it difficult to use the fallback code for module number in `__post_init__`.

src/extra/calibration.py
philsmt: Well spotted, thanks.

tests/test_calibration.py
philsmt: Yes, good point! Thanks

src/extra/calibration.py
takluyver: OK, good enough, then.

tests/test_calibration.py
```diff
@@ -405,12 +400,16 @@ def test_DetectorData_from_identifier():
     assert list(agipd)[0] == next(iter(agipd.keys())) == 'AGIPD00'
     assert isinstance(next(iter(agipd.values())), DetectorModule)
     assert agipd[0] == agipd['AGIPD00']
+    assert agipd.source_names == [
+        f'SPB_DET_AGIPD1M-1/DET/{i}CH0:xtdf'
+        for i, name in range(len(agipd.source_names))]
```
takluyver:
```suggestion
        for i in range(len(agipd.source_names))]
```

takluyver:
LGTM

philsmt:
Hmmm, I had not yet seen your LGTM and thought I had forgot to push. It weirded me out now to see I had, but yet still the push was successful. Didn't see your test fix there!

Thanks for review!

PR #483, opened 2026-04-16
Changed the _fromdict in SerializableMixin to be a class method

TimBerberich:
Separate merge request for the SerializableMixin change.

### DIFF
--- a/src/extra/applications/base.py
+++ b/src/extra/applications/base.py
@@ -63,21 +63,19 @@ class SerializableMixin(object):
         """
         pass
 
-    def _fromdict(self, all_data):
+    @classmethod
+    def _fromdict(cls, all_data):
         """
         Rebuild it from dict.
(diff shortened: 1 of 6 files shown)


src/extra/applications/base.py
```diff
@@ -63,21 +63,19 @@ def _asdict(self):
         """
         pass
 
-    def _fromdict(self, all_data):
+    @classmethod
+    def _fromdict(cls, all_data):
```
daniloefl: Perhaps for clarity explain what the return convention is for this functions and how it should be specialized. This was not written before because I thought it was self-explanatory, but I think in the new setup it is not so obvious.

daniloefl (commented):
As mentioned, only the change log needs to be adapted. Some extra documentation would be good. Otherwise, LGTM!

PR #484, opened 2026-04-16
Add function to find LPD darks with memory cell fallback

takluyver:
When FXE use LPD with different memory cell orders, they would like the option to either take darks with matching memory cell order, or take darks with all memory cells. AIUI, matching the memory cell order gives the best quality, but it's convenient sometimes to have one set of 'good enough' darks while the memory cell pattern changes.

This provides a wrapper around `CalibrationData.from_condition()` to do that. It will use the constant with the memory cell order if that's up to `preference_time` older (more precisely, further in time) than the fallback. I've set this to 5 days by default, but I want to check that with the scientists.

It feels a bit ugly to have a random detector specific lookup function like this, but I haven't thought of a more elegant way to fit it into the API. AFAIK it requires 2 separate requests to get the constants to compare.


src/extra/calibration.py
```diff
@@ -1349,6 +1350,61 @@ def make_dict(self, parameters):
...
+        event_at = datetime.fromisoformat(event_at)
+
+    cd_preferred = CalibrationData.from_condition(
+        condition,
+        detector_name,
+        calibrations=["Offset", "Noise", "BadPixelsDark"],
+        **kwargs
+    )
+
+    fallback_mem_cell_order = ",".join([str(i) for i in range(510)]) + ","
```
philsmt, line 1379: Wouldn't it be possible for `condition` to contain a memory cell count below 510/500, whether with or without memory cell order?

src/extra/calibration.py
```diff
@@ -1349,6 +1350,61 @@ def make_dict(self, parameters):
...
+    if condition.memory_cell_order == fallback_mem_cell_order:
+        return cd_preferred  # Fallback would be the same
+
+    cd_fallback = CalibrationData.from_condition(
+        replace(condition, memory_cell_order=fallback_mem_cell_order),
```
philsmt, line 1384: Please correct me if I'm wrong, but from the MR description, I assumed you want to compare constants with a certain memory cell order with constants that cover the relevant number of memory cells without one. But isn't `memory_cell_order` then empty?

src/extra/calibration.py
```diff
@@ -1349,6 +1350,61 @@ def make_dict(self, parameters):
...
+    # If only one has succeeded, return that one
+    if 'Offset' not in cd_fallback:
+        return cd_preferred
+    if 'Offset' not in cd_preferred:
+        return cd_fallback
+
+    # How far is each set of constants in time? We'll look only at offsets,
+    # which are most significant, and take the median.
+    def time_distance(cd):
+        return np.median([
```
philsmt: This would let arbitrary old CCVs pass as long as it's less than half of them. Mismatching times should be rare one way or the other, but the median also doesn't seem a particularly effective metric here. I would just use the oldest CCV for comparison to avoid an additional loss in data quality across modules.

philsmt (commented):
I was wondering about this discussion recently, and whether there is a clear path to proceed.

I agree it's not pretty, but this is bound to happen given it fundamentally runs against CalCat's design. I will also be hard to implement transparently in `CalibrationData.from_data()` 😢 

I was slightly confused by the implementation though - I assumed the idea is to use in-order constants even in the case of veto with wraparound. The former constants would be saved without a memory cell order condition (rather than all enumerated cells), and IIRC there's no guarantee that in-order constants actually cover all memory cells. As a matter of fact, I think to remember the instrument specifically wanted the flexibility to still take them at any number of memory cells for convenience, whether in-order or not. But naturally the less memory cells are used, the less likely a wraparound becomes.

src/extra/calibration.py
philsmt: Obsolete after discussion in dev team meeting.

src/extra/calibration.py
philsmt: Obsolete after discussion in dev team meeting.

src/extra/calibration.py
philsmt: Discussed this in dev team meeting and decide to determine this for every PDU individually.

takluyver:
To record for future reference: we changed the config for using memory cell order to 'always' about 1 year ago, both for injection and lookup, so recent constants are all injected with memory cell order, whether they wrap around or not (e.g. [this constant](https://in.xfel.eu/calibration/calibration_constants/34431#condition)).

The parameter for the number of memory cells is also left fixed at 512 - there's one counterexample in the config, but it's an early proposal from 2017/2018.

takluyver:
After speaking to Hazem about this, I changed the default preference time to 7 days. This means that a dark taken with a veto pattern at the start of a normal user beamtime will be preferred over darks with no veto (for data with a matching pattern) until the end of that beamtime, which is a convenient way to think about it.

takluyver:
pycalibration draft MR to use this function: https://git.xfel.eu/calibration/pycalibration/-/merge_requests/1238

src/extra/calibration.py
```diff
@@ -1349,6 +1350,68 @@ def make_dict(self, parameters):
...
+
+    cd_preferred = CalibrationData.from_condition(
+        condition,
+        detector_name,
+        calibrations=["Offset", "Noise", "BadPixelsDark"],
+        **kwargs
+    )
+
+    fallback_mem_cell_order = ",".join([str(i) for i in range(510)]) + ","
+    if condition.memory_cell_order == fallback_mem_cell_order:
```
philsmt, line 1380: Given we're now assuming a memory cell order, shall the function reject conditions without one (or an empty one)?

philsmt (commented):
Thanks, LGTM as discussed.

src/extra/calibration.py
takluyver: For interactive use, it's probably quite convenient to allow not specifying a cell order and finding constants with all cells filled in order. On the other hand, it might lead to sloppiness about conditions and further confusion about how pieces interact. :shrug:

philsmt: But does that happen? If I don't pass any memory cell order, `cd_preferred` is retrieved without an order. This condition  will fail, and `cd_preferred` is returned.

takluyver: The shortcut is if this condition *succeeds*, i.e. the fallback wouldn't change `condition`. If it fails, it should carry on and do a second query for the fallback case.

philsmt: Hmmm, not sure how I came to that conclusion. Looks like I read it as an in-equality.

PR #485, opened 2026-04-23
Angular correlation code adde to extra.utils

TimBerberich:
This adds extra.utils.xcca which contains routines to compute cross-correlations of polar data with and without detector mask.
It also contains routines to compute cumulative/(iterative) averages and variances which are useful if you want compute the average ccf by splitting your entire dataset in parts and then merging computed partial results.

Example:
```py
from extra.utils.xcca import AngularCorrelator, AveragedAngularCorrelationMasked
import numpy as np

data = np.random.rand(100,32,64)
mask = np.random.rand(100,32,64)>0.7

ac = AngularCorrelator(*data.shape[1:])

#compute cross-correlation
ccf  = ac.ccf(data[0],mask=mask[0])
#compute harmonic coefficients of cross-correlation
ccn  = ac.ccn(data[0],mask=mask[0])

# compute average correlation of data parts 1 and 2
accn1 = AveragedAngularCorrelationMasked.from_dataset(data[:50],mask[:50])
accn2 = AveragedAngularCorrelationMasked.from_dataset(data[50:],mask[50:])

# combine partial results
accn1.merge(accn2)

# accessing mean and variance
accn1.mean
accn1.variance
```

### DIFF
--- a/docs/applications/index.md
+++ b/docs/applications/index.md
@@ -8,3 +8,4 @@ Applications index:
     - [CookieboxCalibration][extra.applications.CookieboxCalibration]
 - [Grating](grating.md)
     - [Grating2DCalibration][extra.applications.Grating2DCalibration]
+- [XCCA](xcca.md)
--- a/docs/applications/xcca.md
+++ b/docs/applications/xcca.md
@@ -1,0 +1,27 @@
+# X-ray cross-correlation analysis (XCCA)
+This is a toolbox containing classes to easily compute
+the averaged angular cross-correlation.
+
+## Angular cross-correlation
+
+::: extra.applications.xcca.AngularCorrelator
+
+::: extra.applications.xcca.AveragedAngularCorrelation
+	options:
+      inherited_members: false
+
+::: extra.applications.xcca.AveragedAngularCorrelationMasked
+	options:
+      inherited_members: false
+	  
+## Incremental variance
+
+::: extra.applications.xcca._CumulativeVarianceBase
+
+::: extra.applications.xcca.CumulativeVariance
+	options:
+      inherited_members: false
+
+::: extra.applications.xcca.CumulativeVarianceMasked
+	options:
+      inherited_members: false
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -57,6 +57,7 @@ Added:
 - [TOFAnalogResponse][extra.applications.TOFAnalogResponse] can deconvolve the Auger-Meitner
   and photon line if the flag `deconvolve` is set to True.
   summary about a pulse pattern (!435).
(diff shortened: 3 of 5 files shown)


src/extra/utils/xcca.py
```diff
@@ -0,0 +1,797 @@
+import numpy as np
+from typing import Self
+from numpy.typing import NDArray,ArrayLike
+
+class AngularCorrelator:
+    r'''
```
JamesWrigley, line 6: Do the docstrings need to be raw strings?

src/extra/utils/xcca.py
```diff
@@ -0,0 +1,797 @@
...
+        max_order : int|None
+            Maximum considered Fourier series order.
+
+        Returns
+        -------
+        tuple(NDArray[np.float64],NDArray[np.bool])
+            ((n_q,n_q,2*max_order), (n_q,n_q,2*max_order)): (Cross-correlation function $C(q_1,q_2,\phi)$, Mask of cross-correlation function.)
+            If max_order == None the last dimension has size n_phi.
+        
+        """
```
JamesWrigley, line 256: Old docstring?

src/extra/utils/xcca.py
```diff
@@ -0,0 +1,797 @@
...
+            val (NDArray): New data point.
+            mask (NDArray[np.bool]): Mask of the new data point.
+        
+        Returns:
+            CumulativeVarianceMasked: self
+        
+        Examples:
+            ```py
+            import numpy as np
+            from EXtra.utils.xcca import CumulativeVarianceMasked
```
JamesWrigley, line 495: The Python module name is lower-case:
```suggestion
            from extra.utils.xcca import CumulativeVarianceMasked
```

(same for the other docstrings below)

JamesWrigley (changes requested):
Broadly looks fine, not that I'm an expert with XCCA 😛 Some thoughts:
- I would suggest putting `AngularCorrelator` and friends under `applications`, it feels too specific a thing to go in `utils`.
- The tests should pass.
- This should have a changelog entry.

JamesWrigley:
(also, please either rebase or do a squash merge :see_no_evil: )

src/extra/utils/xcca.py
TimBerberich: I typically do this because I have latex code with \ and _ in the docstring, that would throw errors if i do not make them raw strings

src/extra/utils/xcca.py
TimBerberich: good catch !

TimBerberich:
So, everything should be ready to merge now.

JamesWrigley (approved):
Needs a rebase to fix the changelog merge conflict, but otherwise LGTM!

PR #486, opened 2026-05-06
Fix scan.plot_bin_by_steps() uncertainty_method parameter

takluyver:
The parameter only changed the plot label, not the calculation.


JamesWrigley (approved):
Oops 🫣 LGTM!

takluyver:
Thanks for the quick review :+1:

PR #493, opened 2026-05-13
StepTimer: Simple overview of what takes time

takluyver:
I want to get a basic idea of what steps in a slow operation are taking time. Profilers can give much more detail, but also require more effort to use. The timer calls added here can be left in functions and methods (`perf_counter()` takes <100 ns), and you can look at the timings from the last call that used them:

image

You instantiate `StepTimer` inside a function with a name for what it's timing, and then call it at the end of a step with a label for that step. Steps with the same label are displayed as repeats of the same step. The finish time is set when the timer goes out of scope.

You can also set `StepTimer.show = True` to print timings as they are collected, or replace the `on_(start|step|finish)` methods to e.g. send it to a log file or a database.

### DIFF
--- a/src/extra/utils/misc.py
+++ b/src/extra/utils/misc.py
@@ -1,8 +1,13 @@
 import sys
+from dataclasses import dataclass
+from time import perf_counter
+from typing import ClassVar, Self
 
 import numpy as np
 
 from typing import Any
+
+from pandas._libs.tslibs import timestamps
 
 
 def find_nearest_index(array, value: Any) -> np.int64:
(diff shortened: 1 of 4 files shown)


src/extra/utils/misc.py
```diff
@@ -58,6 +63,66 @@ def reorder_axes_to_shape(a, target_shape):
...
+    def step_times(self):
+        d = {}
+        prev_ts = self.timestamps[0][1]
+        for label, ts in self.timestamps[1:]:
+            d.setdefault(label, []).append(ts - prev_ts)
+            prev_ts = ts
+
+        return d
+
+    def _repr_markdown_(self):
```
daniloefl, line 81: I like this! But why not also overwrite `__repr__`, so it would also be seen if we try to print it in the REPL? This may not be the main use-case, but I can imagine debugging this inside something else not running in Jupyter and it could be useful.

src/extra/utils/misc.py
```diff
@@ -58,6 +63,66 @@ def reorder_axes_to_shape(a, target_shape):
...
+        lines = [f"**{self.name}** step timings:", ""]
+        for i, (label, dts) in enumerate(self.step_times.items(), start=1):
+            dt = np.mean(dts)
+            rpt = f"({len(dts)}×)" if len(dts) > 1 else ""
+            lines.append(f"{i}. {label}: {dt:.3g} s {rpt}")
```
daniloefl: Would it be useful to do `{i:3d}` and similarly for other entries to align the outputs, such that it is easy to see them in a table.

src/extra/utils/misc.py
```diff
@@ -58,6 +63,66 @@ def reorder_axes_to_shape(a, target_shape):
...
+        if self.finish:
+            start = self.timestamps[0][1]
+            lines.extend(["", f"Total: {self.finish - start:.3g} s"])
+        return "\n".join(lines)
+
+
+class StepTimer:
```
daniloefl, line 109: I know it is silly to ask for this for such an auxiliary task, but what about adding some docstrings to guide others?

src/extra/utils/misc.py
```diff
@@ -58,6 +63,66 @@ def reorder_axes_to_shape(a, target_shape):
     return a.transpose(order)
 
 
+@dataclass
+class TimingResult:
```
daniloefl, line 65: What about some docstrings?

daniloefl (commented):
This is very good and useful as it is. I have no major comments, other than nitpicking. 
I mentioned docstrings, but it really is not a major concern here.

Feel free to take my comments or leave things as they are.

src/extra/utils/misc.py
takluyver: I generally avoid multi-line reprs, but I guess this is somewhere that it might be reasonable to make an exception.

src/extra/utils/misc.py
takluyver: The markdown is rendered in a variable-width font, so giving lines a set number of characters won't align them. But we could do a table - possibly I was just being lazy

takluyver: Here's what it could look like as a table in Markdown

image


And here it is with a table with no header row (in HTML):

image

src/extra/utils/misc.py
takluyver: Not silly at all, docstrings make sense

philsmt:
Do you have any thoughts about making this a context manager, automatically adding the total runtime as a step?

takluyver:
It measures to when it's deleted - i.e. normally when the function it's in returns. I wanted to avoid needing an extra level of indentation for `with` blocks everywhere.

There's no reason it couldn't be a context manager as well, though. So you'd record a step when the block exits, maybe labelled `exit` or `__exit__`? Would you also record block entry, or assume that it's always created and entered at the same time, as in:

```python
with StepTimer("foo") as timer:
    ...
```

philsmt:
My personal motivation was actually just to time the duration of a block. I like context managers for this purpose, as it's easy to quickly indent a block and the context manager up-front - however I agree it can be cumbersome in production code.

Combined with the more advanced functionality of `StepTimer`, yes, essentially the context manager could be one additional step with a fixed name, and you can use the `timer` object during the block as well for additional steps.

src/extra/utils/misc.py
daniloefl: No, you are right: my comment makes no sense if this is rendered in Markdown. I only did not think of it, because I usually see "|" to split columns in Markdown. I am not sure how you avoid the "|", but in that case, ignore it.

takluyver:
I've changed the Markdown format to a table, showing the total time on each step as well as the stdev.

image

daniloefl:
Looks great to me after all the changes!

takluyver:
Thanks!

PR #497, opened 2026-05-26
Parallelize AdqRawChannel using pasha

philsmt:
This branch has grown stale after some unusual scaling was found with it that led to discussions DESY folks to improve. Recently Danilo has stumbled upon this to improve performance, and kindly redone some tests (stealing his results here):
image
This scaling could be seen for both GPFS and dCache, so let's get this in!

### DIFF
--- a/src/extra/components/adq.py
+++ b/src/extra/components/adq.py
@@ -328,13 +328,13 @@ class AdqRawChannel:
 
         return edge_func
 
-    def _validate_out(self, out, shape):
+    def _validate_out(self, out, shape, alloc=np.zeros):
         """Validate output arguments."""
 
         is_corrected = self._cm_period > 0 or self._baselevel is not None
 
         if out is None:
-            out = np.empty(shape, dtype=np.float32)
+            out = alloc(shape, dtype=np.float32)
         elif any([a < b for a, b in zip(out.shape, shape)]):
             raise ValueError(f'requires at least output array shape {shape}')
         elif is_corrected and not np.issubdtype(out.dtype, np.floating):
@@ -908,7 +908,7 @@ class AdqRawChannel:
                 on each train trace, extra.signal.dled by default.
             max_edges (int, optional): Maximal number of edges per
                 train, 50 by default.
-            parallel (int or None, optional): Nunmber of parallel
+            parallel (int or None, optional): Number of parallel
                 processes to use, by default 10 or a quarter of all cores
                 whichever is lower. Any non-positive value or 1 disable
                 parallelization.
@@ -952,7 +952,7 @@ class AdqRawChannel:
                 on each train trace, extra.signal.dled by default.
             max_edges (int, optional): Maximal number of edges per
                 train, 1/5000 of trace length by default.
-            parallel (int or None, optional): Nunmber of parallel
+            parallel (int or None, optional): Number of parallel
                 processes to use, by default 10 or a quarter of all cores
                 whichever is lower. Any non-positive value or 1 disable
                 parallelization.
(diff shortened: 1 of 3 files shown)


src/extra/components/adq.py
daniloefl, line 1027: Do you think it makes sense to document the new argument in the docstring?

I think it is a bit confusing that parallel indicates the degree of paralellization (as a user I would naively assume that I should do `parallel=True` to parallelize). If this is the standard, that is ok, but I would then add comments on the expected behaviour as docstrings.

src/extra/components/adq.py
```diff
@@ -1006,7 +1006,7 @@ def digitize_edges(wid, index, trace):
         return self._shape_edge_array(edges, amplitudes, orig_coords,
                                       labelled, squeeze_edges)
 
-    def train_data(self, labelled=True, roi=(), out=None):
+    def train_data(self, labelled=True, roi=(), out=None, parallel=None):
```
daniloefl, line 1009: I thought the sentence in the change log says it runs in parallel by default, but it seems that `parallel=None` by default. Is this really the correct default argument value?

src/extra/components/adq.py
```diff
@@ -1062,7 +1075,7 @@ def train_data(self, labelled=True, roi=(), out=None):
         return xr.DataArray(out, coords=coords)
 
     def pulse_data(self, labelled=True, pulse_dim='pulseId', train_roi=(),
```
daniloefl, line 1081: As in the previous comment, following the change log changes, shouldn't parallel be `!= None` by default?

src/extra/components/adq.py
```diff
@@ -1102,25 +1115,47 @@ def pulse_data(self, labelled=True, pulse_dim='pulseId', train_roi=(),
...
 
-        # Prepare output buffer.
+        # Prepare output buffer shape.
         out_shape = (len(pulse_ids), pulse_layout['length'].max())
-        out = self._validate_out(out, out_shape)
 
-        # Temporary buffer for a single iteration.
-        tmp = np.zeros((200,) + roi_shape(
-            self._raw_key.entry_shape, train_roi), dtype=out.dtype)
+        if parallel is not False:
```
daniloefl, line 1129: And here, different from `train_data`, there is a check for `False` instead of `None`. Is this on purpose?

src/extra/components/adq.py
daniloefl, line 1104: As before, shouldn't there be a docstring about `parallel`?

src/extra/components/adq.py
daniloefl: PS: I see this is tested in the mock test, but perhaps it would be useful to be consistent for code readability.

daniloefl (commented):
This looks already good to me. I have only added a few small questions about documentation and readability of the code, but feel free to ignore it all.

src/extra/components/adq.py
philsmt: Well spotted, I forgot about the docstring.

The use of `parallel` in this manner is both established in this component and others, yes. The underlying idea is to have it enabled by default with auto values, but allow fine tuning or disabling it. The canonical value in Python for default value is `None`. Note that `True` is indeed supported in case a mistake is made and treated as `None` in `_prepare_pasha()`, but not documented to not recommend its use.

src/extra/components/adq.py
philsmt: As mentioned above, `None` is the default value and thus parallel by default. `None` does not mean "parallelization is none", but the absence of a value for `parallel`.

src/extra/components/adq.py
philsmt: see above.

src/extra/components/adq.py
philsmt: Well spotted, thanks! It is meant to be `is not False` as in all other cases.

src/extra/components/adq.py
philsmt: certainly, fixed with the above.

src/extra/components/adq.py
daniloefl: Understood. Thank you.

src/extra/components/adq.py
daniloefl: Thank you.

daniloefl:
Thank you for the changes! Looks good!

philsmt:
Thanks for the prompt review!

PR #498, opened 2026-05-27
Bug fix: number of photo-electron counts.

daniloefl:
This is reminiscent of a previous PR, in which the variable names were adapted, but it seems this variable name was not changed.


philsmt:
LGTM

PR #499, opened 2026-05-28
Add TimeserverPulses.plot_xray_patterns()

philsmt:
Adds a new method `TimeserverPulses.plot_xray_patterns()` that generates a visualized summary of the X-ray pulse patterns in this data, including support for variable patterns using weighted transparency:

image

Given this data should be available for all supported data formats, it can be generated from any `PulsePattern` object.

### DIFF
--- a/src/extra/components/pulses.py
+++ b/src/extra/components/pulses.py
@@ -931,6 +931,14 @@ class TimeserverPulses(PulsePattern):
         """Get nodes in pulse pattern decoder device."""
         raise NotImplementedError('_get_ppdecoder_nodes')
 
+    def _get_xray_pulses(self, sase):
+        """XrayPulses object for given SASE using the same data."""
+
+        if isinstance(self, XrayPulses) and self._sase == sase:
+            return self
+
+        return XrayPulses(None, self._source, sase=sase)
+
     @property
     def timeserver(self) -> SourceData:
         """Used timeserver source."""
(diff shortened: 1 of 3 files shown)


src/extra/components/pulses.py
```diff
@@ -931,6 +931,12 @@ def _get_ppdecoder_nodes(self):
         """Get nodes in pulse pattern decoder device."""
         raise NotImplementedError('_get_ppdecoder_nodes')
 
+    def _get_xray_pulses(self, sase):
```
daniloefl, line 934: What about a one-liner docstring?

src/extra/components/pulses.py
```diff
@@ -998,12 +1004,78 @@ def is_sa1_interleaved_with_sa3(self) -> Optional[bool]:
...
+    def plot_xray_patterns(self, figsize=(9, 4), ax=None):
+        """Visualize X-ray pulse patterns.
+
+        Plots the X-ray pulse pattterns of SA1, SA2 and SA3 in this
+        data on a common time scale. Variable patterns are represented
+        by levels of transparency weighted by their rate of occurence.
 
-        return sa1_pulses.is_interleaved_with(sa3_pulses)
+        Arguments:
+            figsize (2-tuple of int): Figure size in inches passed to
```
daniloefl: Does it have to be int? I thought those were inches.

daniloefl (commented):
This looks quite good! But I did not understand why the y spread of the bar in the plot covers multiple SASEs. Is that on purpose?

fadybishara:
> This looks quite good! But I did not understand why the y spread of the bar in the plot covers multiple SASEs. Is that on purpose?

Probably/maybe because it would look better if SASE1 and SASE3 are interleaved?

src/extra/components/pulses.py
```diff
@@ -998,12 +1004,78 @@ def is_sa1_interleaved_with_sa3(self) -> Optional[bool]:
...
+        pid_range = max_pid - min_pid
+        ax.set_xlim(min_pid - pid_range*0.05, max_pid + pid_range*0.05)
+        ax.set_xlabel('Pulse ID')
+
+        from matplotlib.ticker import AutoLocator
+        t_ax = ax.twiny()
+        t_ax.set_xlim(ax.get_xlim())
+
+        # Compute suitable ticks in time based on the pulse ID range.
+        ms_rate = self.bunch_repetition_rate * 1e-9
```
fadybishara: Ah! Here is the problem. `self.bunch_repetition_rate` is in Hz so this should be
```suggestion
        ms_rate = 1 / self.bunch_repetition_rate * 1e3
```
This way the units work out :
```math
\mathrm{s} \times \frac{10^3 \mathrm{ms}}{\mathrm{s}}
```
But in this case, maybe it's better to do μs instead of ms, no?

src/extra/components/pulses.py
philsmt: It seemed pretty obvious as a private method, but fair point that all other privates also have one.

src/extra/components/pulses.py
philsmt: Yes, certainly!

philsmt:
> > This looks quite good! But I did not understand why the y spread of the bar in the plot covers multiple SASEs. Is that on purpose?
> 
> Probably/maybe because it would look better if SASE1 and SASE3 are interleaved?

Fady answered the question perfectly. I tried to make a plot to remains as readable as possible in all kind of situations of variable and/or interleaved patterns. The shift in Y should always give some space to see a SASE's pattern isolated.

src/extra/components/pulses.py
philsmt: As mentioned in Zulip, I messed up conversion when changing to `self.bunch_repetition_rate`. And yeah, I think microseconds looks better.

daniloefl:
I see. As I understood it at the moment this y axis behaviour is always done, regardless if the pulses are interleaved or not (specifically, I don't see a check of `is_sa1_interleaved_with_sa3()` in the plot code, but I may simply have missed it). Is this done somewhere? Or is this y-axis displacement always like this.

This is really not important, so feel free to go ahead as is.

src/extra/components/pulses.py
fadybishara: Sorry, I don't follow, why should figsize be a 2-tuple of ints? 
[matplotlib.pyplot.figure](https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.figure.html#matplotlib.pyplot.figure) expects a 2-tuple of floats...

Did I misunderstand what you meant by your answer?

philsmt: I meant to answer to Danilo's last sentence, it should be inches and thus floats.

philsmt:
Hmmm, I would prefer to have the same layout in all situations whether they are clearly separated visually or not.

daniloefl:
Then just ignore my comment. I was going to suggest to remove the overlap, as it conveys no meaning in this case and it can make someone think they are interleaved when they are not, but if you think this is not a risk, it does look pretty.

tmichela:
Alternative idea, I quite like how we visualize patterns in the pulse ID map / lit frame finder Karabo gui widget

pulseid_map

philsmt:
> Alternative idea, I quite like how we visualize patterns in the pulse ID map / lit frame finder Karabo gui widget

I'm planning to use that for visualization of a `PulsePattern`'s own pulse IDs.

It's fine for the entire pattern, but I doesn't seem to me its benefits are what you're looking for there. The grid is convenient for counting your pattern (evern nth pulse), but not so much for time distance. It would also be more complicated with overlapping pulses. Hence I went for a line plot for the entire pattern across SASEs, and will try the grid for own pulse IDs. Your example is actually such a case, and fancy types that combine Xray and other pulses could customize the grid to plot just that.

philsmt:
I've added a couple of aesthetic and technical improvements without any fundamental change. I will merge this by the end of the day unless there are objections. You can find the current version in the screenshot above.

fadybishara:
Had a quick face-to-face discussion with @philsmt and worked out a couple of aesthetic tweaks. LGTM!

philsmt:
In addition to Fady's suggestions, I added a right Y axis with pulse statistics:

image

PR #501, opened 2026-05-29
Fix incomplete pulse alignment in new parallel paths of AdqRawChannel

philsmt:
Merging yesterday's MR https://github.com/European-XFEL/EXtra/pull/497 unfortunately introduces the already fixed alignment issues from https://github.com/European-XFEL/EXtra/pull/361 and https://github.com/European-XFEL/EXtra/pull/332. I'm still a bit mystified why the tests existing for this particular issue did not trigger, but I would like to have it fixed ASAP.

### DIFF
--- a/src/extra/components/adq.py
+++ b/src/extra/components/adq.py
@@ -1056,7 +1056,7 @@ class AdqRawChannel:
                 out[index] = data[roi]
                 self._preprocess(out[index], out[index])
 
-            psh.map(read_train, self._raw_key)
+            psh.map(read_train, self._raw_key.drop_empty_trains())
         else:
             out = self._validate_out(out, shape)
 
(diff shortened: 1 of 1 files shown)


src/extra/components/adq.py
daniloefl, line 1064: Just out of curiosity and excess of caution: why no `drop_empty_trains()` here?

daniloefl: I can see this is done implicitly (through doing it in `raw_key` instead of `self._raw_key` in line 1156, for instance.

daniloefl (commented):
Looks good. I only have a small question for my own education. I am not sure why `drop_empty_trains()` is not done in line 1064. Otherwise, go ahead.

src/extra/components/adq.py
philsmt: Because all access done there only takes trains with data into account, specifically `KeyData.shape` and `KeyData.ndarray()`:
```
                num_trains = int(kd.shape[0])
                out_sel = out[offset:offset+num_trains]
                offset += num_trains

                kd.ndarray(roi=roi, out=out_sel)  
```

The issue stems from `pasha.map` iterating over all selected trains (and passing `None` if no data is present) rather than all trains with data.

daniloefl: I see. Thank you!

daniloefl:
With that one thing understood (by me), feel free to merge it.

philsmt:
Thanks for the quick look!

PR #502, opened 2026-06-01
Fix lpd_dark_consts_with_fallback with event_at parameter

takluyver:
The `event_at` parameter was not being passed through to querying CalCat. This was causing test failures on Gitlab (where it doesn't use the recorded responses), because it was picking up a newer constant.

cc @philsmt


philsmt:
While we're at it, how about a `client` argument?

takluyver:
It passes through arbitrary `**kwargs` to `from_condition`, so passing `client=` should already work. But this didn't apply to `event_at` since I'd made it a separate parameter.

philsmt:
Ah, fair point. LGTM

PR #503, opened 2026-06-01
AdqRawChannel improvements

takluyver:
- Allow specifying a dtype for `pulse_data` without specifying an entire output array - it needs a non-trivial amount of code to work out the correct output shape. We're disabling the common-mode correction, so keeping the data as int16 is faster than converting to float.
  - I wondered about making int16 the default when you're not using common-mode or baselevel, but I'm not convinced it's worth the API break.
  - We could also add the parameter for `train_data`? But it's not so important because the shape is easier to work out.
- Add a `.split_trains()` method to split the AdqRawChannel into chunks - this is something we want because we reduce the data immediately after loading, so doing it piece by piece avoids a big spike in memory usage.

### DIFF
--- a/src/extra/components/adq.py
+++ b/src/extra/components/adq.py
@@ -1054,8 +1054,11 @@ class AdqRawChannel:
 
         if parallel is not False:
             # Prepare parallelization.
+            import pasha
             psh = self._prepare_pasha(parallel)
 
+            if isinstance(psh, pasha.ProcessContext) and out is not None:
+                raise TypeError("Cannot use out= with parallel processing")
             out = self._validate_out(out, shape, psh.alloc)
 
             def read_train(wid, index, train_id, data):
@@ -1136,12 +1139,13 @@ class AdqRawChannel:
 
         if parallel is not False:
             # Prepare parallelization.
+            import pasha
             psh = self._prepare_pasha(parallel)
 
             # Prepare output buffers.
-            if out is not None:
-                raise ValueError("Cannot use output array with parallel processing")
-            out = self._validate_out(None, out_shape, psh.alloc, dtype=dtype)
+            if isinstance(psh, pasha.ProcessContext) and out is not None:
+                raise TypeError("Cannot use out= with parallel processing")
+            out = self._validate_out(out, out_shape, psh.alloc, dtype=dtype)
 
             def read_pulses(wid, index, train_id, data):
                 layout_row = pulse_layout.loc[train_id]
--- a/src/extra/components/adq.py
+++ b/src/extra/components/adq.py
@@ -328,13 +328,13 @@ class AdqRawChannel:
 
         return edge_func
 
-    def _validate_out(self, out, shape, alloc=np.zeros):
+    def _validate_out(self, out, shape, alloc=np.zeros, dtype=np.float32):
         """Validate output arguments."""
 
         is_corrected = self._cm_period > 0 or self._baselevel is not None
 
         if out is None:
-            out = alloc(shape, dtype=np.float32)
+            out = alloc(shape, dtype=dtype)
         elif any([a < b for a, b in zip(out.shape, shape)]):
             raise ValueError(f'requires at least output array shape {shape}')
         elif is_corrected and not np.issubdtype(out.dtype, np.floating):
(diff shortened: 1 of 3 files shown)


src/extra/components/adq.py
```diff
@@ -1131,7 +1139,9 @@ def pulse_data(self, labelled=True, pulse_dim='pulseId', train_roi=(),
             psh = self._prepare_pasha(parallel)
 
             # Prepare output buffers.
-            out = self._validate_out(None, out_shape, psh.alloc)
+            if out is not None:
+                raise ValueError("Cannot use output array with parallel processing")
```
takluyver: I think it shouldn't just silently ignore `out`. However, `train_data` attempts to use it even in the parallel case. We should make them consistent, whatever we prefer.

philsmt: If I see this correctly, it's even breaking right now with passing `out`, I completely overlooked this.

Ideally we would be able to pass something allocated by `psh.alloc` and identify this here, but I'm not sure of a good way. So this is definitely safer ofr now.

philsmt (approved):
LGTM

src/extra/components/adq.py
takluyver: Thanks. I don't know of a good way to identify if an array is in shared memory either, so for now I'll add the same check in `.train_data()`

takluyver:
I realised both of these methods have 2 ways to run things in serial - passing `parallel=1` (or lower) uses a pasha SerialContext, and `parallel=False` (undocumented) falls back to the older code, which loads up to 200 trains at a time.

It looks like the digitizer data at FXE is recorded in chunks of 100 trains, so I suspect reading a block of trains together is going to be a lot more efficient.

src/extra/components/adq.py
philsmt: The best I can think of would be that `pasha` holds weak references to any array it allocated?

takluyver:
> I suspect reading a block of trains together is going to be a lot more efficient.

This guess seems to be wrong. I tested `.train_data()` with 1000 trains of data, disabling common mode to focus on loading performance. Passing `parallel=False`  it takes about 1.0s, whereas with `parallel=1` it takes about 0.73 s. This seems to be reproducible. Using a simple loop over `.trains()` is similar to the SerialContext case. The results are also similar for `pulse_data()`, just slightly slower.

I'm a bit confused by how this can work - I was expecting HDF5 to read in the whole ~160 MB chunk again on each access, which would make it very slow. I guess it must have got smarter - maybe it will keep 1 chunk in the cache even if that exceeds the chunk size.

However, in this scenario, parallelism actually makes it slower: ~1.8 seconds for the same 1000 trains. I'm not sure why, maybe the extra cost of dealing with shared memory.

philsmt:
😮‍💨 

So we're back to parallel reading not always being faster.

takluyver:
I'm afraid so.

Also, when I use `dtype=` to avoid the type conversion, reading bigger chunks (`parallel=False`) is a bit faster than iterating train by train, although not by much.

image

src/extra/components/adq.py
takluyver: I think that would probably work well enough - in theory you could make an array in shared memory outside of pasha, but that's probably not an important use case to handle. You'd want to check against `arr.base` as well to catch views made from pasha-allocated arrays

philsmt:
I suspect all these numbers are for dCache?

takluyver:
GPFS, I think. I was testing with p11696 r112 - it looks like raw data from that proposal is still on GPFS.

src/extra/components/adq.py
```diff
@@ -1048,8 +1054,11 @@ def train_data(self, labelled=True, roi=(), out=None, parallel=None):
 
         if parallel is not False:
             # Prepare parallelization.
+            import pasha
             psh = self._prepare_pasha(parallel)
 
+            if isinstance(psh, pasha.ProcessContext) and out is not None:
+                raise TypeError("Cannot use out= with parallel processing")
```
philsmt, line 1061: You think it's clear from this exception `ThreadContext` works? (whether's useful or not)

takluyver: As things stand, there's no way for it to get ThreadContext - `_prepare_pasha()` will either create a ProcessContext or a SerialContext. And the pasha objects are an internal detail - the API just takes `parallel=`. So I think this is OK.

PR #505, opened 2026-06-04
Add amplitude support to DelayLineDetector.edges()

philsmt:
For a while now, reconstructed DLD data also includes the amplitudes of discriminated edges in the `raw.amplitudes` key. This PR adds support to `DelayLineDetector.edges()` to include this data alongside edge positions, if requested.


philsmt:
(merging without further review for this week's beamtime)

PR #506, opened 2026-06-05
Fixed behaviour of TOF response reading due to change in API in PR483.

daniloefl:
PR #483 changed the API for the serialization and this caused a bug in `TOFAnalogResponse`. This PR fixes that and additionally makes a small change in the order of operations for data reading to take advantage of the parallelization implemented in PR #497 .


TimBerberich:
LGTM, go ahead and merge

daniloefl:
Thank you! I just added a few more tests to increase the coverage due to the change. This was good, as it lead to one bug being caught and fixed before the merge.

PR #507, opened 2026-06-05
Split extra.calibration into its own package

philsmt:
Over the past few months, the `extra.calibration` module has grown significantly in size to well over 2000 lines. This PR splits up the module into its own package into orthogonal parts:

- `__init__.py` containing `BadPixels` and prior imports
- `calcat.py` containing the CalCat interface and infrastructure handling
- `detector.py` containing the `DetectorData` API
- `conditions.py` containing the condition dataclasses
- `constants.py` containing `CalibrationData` API

This change should be entirely transparent to users unless I've missed importing some public symbol into the package or use of a private symbol.


takluyver:
I haven't gone through the diff closely, but I don't think that's necessary. This makes sense to me.

philsmt:
Merging this by the end of this week.

PR #508, opened 2026-06-09
Add split_trains() method to more component classes

takluyver:
This adds select_trains() to `Timepix3` and `XGM`, and split_trains() to those plus `DelayLineDetector` and the pulse pattern components.

While running the tests on this, I discovered that generating the DLD mock data for tests was very slow, due to the way h5py does broadcasting, so the first commit is a workaround for that.

### DIFF
--- a/src/extra/components/pulses.py
+++ b/src/extra/components/pulses.py
@@ -16,6 +16,7 @@ from euxfel_bunch_pattern import is_sase, is_laser, \
     PPL_BITS, DESTINATION_TLD, DESTINATION_T4D, DESTINATION_T5D, \
     PHOTON_LINE_DEFLECTION
 from extra_data import DataCollection, SourceData, KeyData, by_index, by_id
+from extra_data.read_machinery import split_trains
 
 from .utils import identify_sase, _instrument_to_sase
 
@@ -359,6 +360,18 @@ class PulsePattern:
             res._pulse_ids = None
 
         return res
+
+    def split_trains(self, parts=None, trains_per_part=None):
+        # Get train IDs from an EXtra-data object if possible, so it will split
+        # the same way as other things from the same data.
+        if self._source is not None:
+            tids = self._source.train_ids
+        elif self._key is not None:
+            tids = self._key.train_ids
+        else:
+            tids = self._get_train_ids()  # Used for ManualPulses
+        for sl in split_trains(len(tids), parts=parts, trains_per_part=trains_per_part):
+            yield self.select_trains(sl)
 
     def select_pulses(self, pulse_sel, train_sel=None,
                       reset_index=True) -> ManualPulses:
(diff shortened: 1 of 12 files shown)


takluyver:
I've added basic tests, but it's not much more than a smoketest, I don't check accessing any data from the split parts.

src/extra/components/pulses.py
```diff
@@ -360,6 +361,18 @@ def select_trains(self, trains):
...
 
+    def split_trains(self, parts=None, trains_per_part=None):
+        # Get train IDs from an EXtra-data object if possible, so it will split
+        # the same way as other things from the same data.
+        if self._source is not None:
+            tids = self._source.train_ids
+        elif self._key is not None:
+            tids = self._key.train_ids
+        else:
+            tids = self._get_train_ids()  # Used for ManualPulses
```
philsmt, line 372: Hmmm, I wanted to comment that `_get_train_ids()` *should* always you the right answer here, but checking the implementations I noticed they're actually doing it wrong 😕. I would like to fix this, as their current behaviour is not useful.

Therefore let's keep this as is for the moment.

src/extra/components/timepix.py
```diff
@@ -108,6 +111,36 @@ def __repr__(self):
         return "<{} {}: {}>".format(type(self).__name__, self._detector_name,
                                     ', '.join(data_labels))
 
+    def select_trains(self, trains):
```
philsmt, line 115: Could this use `extra.components.utils._select_subcomponent_trains` like [here](https://github.com/European-XFEL/EXtra/blob/master/src/extra/components/las.py#L285)?

philsmt (approved):
Thanks, only a minor comment. However, I did realize most implementations in `PulsePattern` are not quite doing what they were meant to, but I will look at that separately.

General comment: It's *only* 3 lines, but copying it around still seems annoying. Would it make sense to generalize it and call that in every component?

src/extra/components/timepix.py
takluyver: Good point, I'd forgotten about that helper function. Although, looking at it, I realise we've never actually tested it, because it has two bugs that would prevent it ever working. :melting_face:

takluyver:
> Would it make sense to generalize it and call that in every component?

I suspect it would end up being 3 lines even with a helper function, by the time you've figured out where the train IDs are coming from and made a nicely formatted function call with keyword arguments. Copying around 3 lines has the advantage that if there are any optimisations possible for a particular class, e.g. to avoid recomputing things on each chunk, it's obvious that they can be added in directly. This isn't a big deal, though, so I'm happy to do another helper if you prefer.

philsmt:
LGTM 🫣

takluyver:
Thank-you!

PR #509, opened 2026-06-12
Speed up CookieboxCalibration and TOFAnalogResponse and use `AdqRawChannel` parallelization when possible

daniloefl:
The applications that depend on `AdqRawChannel` have now switched to use the inherent parallelization there. This simplifies the software, makes it more maintainable and achieves higher speeds, as I/O is the main reason for the slow processing.

Additionally, this speeds up the processing further by:
1. Using `np.intersect1d` which is faster than using `intersect` on sets.
2. Only calculating the mean pulse energy if a selection is applied.

Other changes using standard maps instead of `ProcessPoolExecutor` remove parallel processing from this code, as it could then compete with the parallel processing in `AdqRawChannel`.

### DIFF
--- a/src/extra/applications/cookiebox.py
+++ b/src/extra/applications/cookiebox.py
@@ -138,7 +138,9 @@ def calc_mean(itr: Tuple[int, int], scan: Scan, xgm_data: xr.DataArray, tof: Dic
               xgm_threshold: float,
               count_threshold: float=None,
               correction_fn=None,
-              count_samples: int=500) -> xr.DataArray:
+              count_samples: int=500,
+              parallel: int=10,
+              ) -> xr.DataArray:
     """
     Calculate the mean of the ToF data in the given tof and energy bin in `itr`.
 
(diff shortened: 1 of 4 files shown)


src/extra/applications/cookiebox.py
```diff
@@ -138,7 +138,9 @@ def calc_mean(itr: Tuple[int, int], scan: Scan, xgm_data: xr.DataArray, tof: Dic
               xgm_threshold: float,
               count_threshold: float=None,
               correction_fn=None,
-              count_samples: int=500) -> xr.DataArray:
+              count_samples: int=500,
+              parallel: int=10,
```
philsmt: Why chose an explicit number of processes here rather than leave it up to `AdqRawChannel` as before? This may run on many different kinds of nodes.

philsmt (approved):
This MR is rather noisy with a major part of the deal unrelated to parallelization, but those parts seem to be alright. LGTM

daniloefl:
> This MR is rather noisy with a major part of the deal unrelated to parallelization, but those parts seem to be alright. LGTM

For the record, there are indeed changes in the way the pulse energy is treated and rewriting of some of the code, but this is also done with the goal of speeding things up, by:
1. Using `np.intersect1d` which is faster than using `instersect` on sets.
2. Only calculating them if a selection over them was required.

Other changes using standard maps instead of `ProcessPoolExecutor` remove parallel processing from this code, as it could then compete with the parallel processing in `AdqRawChannel`.

I will edit the description to include this information.

src/extra/applications/cookiebox.py
daniloefl: This is true. I have tackled this by setting it to `None` by default. Additionally, `parallel` is now only set in functions that use it, and not serialized to HDF5.

PR #510, opened 2026-06-23
Improve robustness of the grating calibration

daniloefl:
This improves the robustness of the grating calibration, by:
- Applying the calibration mask, if provided; and
- Smoothing the data if required for the purposes of obtaining the calibration function.

The data is interpolated for the masked points, getting rid of fluctuations caused by it.

The background subtraction is also removed, so that we rely instead on the calibration framework.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -73,7 +73,8 @@ Added:
   and photon line if the flag `deconvolve` is set to True.
   summary about a pulse pattern (!435).
 - `StepTimer` class for measuring performance within functions.
-- [Grating1DCalibration][extra.applications.Grating1DCalibration] can use the calibration mask
+- [Grating1DCalibration][extra.applications.Grating1DCalibration] can use the calibration mask,
+  removes the background subtraction (leaving it to the calibration DB),
   and use robust tools to improve the fit quality (!510).
 
 Fixed:
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -73,6 +73,8 @@ Added:
   and photon line if the flag `deconvolve` is set to True.
   summary about a pulse pattern (!435).
 - `StepTimer` class for measuring performance within functions.
+- [Grating1DCalibration][extra.applications.Grating1DCalibration] can use the calibration mask
+  and use robust tools to improve the fit quality (!510).
 
 Fixed:
 
--- a/src/extra/applications/grating.py
+++ b/src/extra/applications/grating.py
@@ -295,7 +295,7 @@ class Grating1DCalibration(SerializableMixin):
 
         Args:
           run: Input run.
-          load_all: If True, load all data in memory at once. This is faste, but uses more memory.
+          load_all: If True, load all data in memory at once. This is faster, but uses more memory.
                     Disable if not enough memory is available.
         """
         # do it per train to avoid memory overflow
@@ -305,7 +305,7 @@ class Grating1DCalibration(SerializableMixin):
             trainId = out_data.trainId.to_numpy()
             out_data = out_data.to_numpy()
             out_data = out_data[:, self.offset::pulse_period, self.min_pixel:self.max_pixel]
-            if len(self.grating_mask_key) > 0:
+            if self.grating_mask_key:
                 out_mask = run[self.grating_source, self.grating_mask_key].ndarray() > 0
                 out_mask = out_mask[:, self.offset::pulse_period, self.min_pixel:self.max_pixel]
                 out_data[out_mask] = np.nan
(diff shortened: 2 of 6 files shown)


src/extra/applications/grating.py
```diff
@@ -23,41 +26,46 @@ def calc_mean(energy_id: int, scan: Scan,
...
-    grating_calib.setup(grating_signal, grating_bkg, scan, pulses)
+    grating_calib.setup(grating_signal, scan, pulses)
 
     # apply it in new data
     new_run = open_run(...)
     grating_calib.apply(new_run)
     ```
     """
-    def __init__(self, offset: Optional[int]=None, min_pixel: int=500, max_pixel: int=1000):
+    def __init__(self, offset: Optional[int]=None, min_pixel: int=0, max_pixel: int=1200, sigma: float=2.0):
```
tmichela: I assume you're the only user of this? as changing default can break existing workflows.

src/extra/applications/grating.py
```diff
@@ -158,9 +164,25 @@ def guess_offset(self):
...
         Guess offset.
         """
-        I = self._grating_signal.xarray().sel(dim_1=np.s_[self.min_pixel:self.max_pixel]).mean('dim_1').mean('trainId')
-        threshold = np.median(I)
-        self.offset = np.where(I >= threshold)[0][0]
+        I = self._grating_signal.xarray().sel(dim_1=np.s_[self.min_pixel:self.max_pixel]).to_numpy()
+        if self._grating_mask is not None:
+            Im = self._grating_mask.xarray().sel(dim_1=np.s_[self.min_pixel:self.max_pixel]).to_numpy()
+            Im = (Im > 0)
+            I[Im] = 0
```
tmichela: Nitpicking, I guess this could be just one line

```suggestion
            I[Im > 0] = 0
```

docs/changelog.md
```diff
@@ -73,6 +73,8 @@ Added:
   and photon line if the flag `deconvolve` is set to True.
   summary about a pulse pattern (!435).
 - `StepTimer` class for measuring performance within functions.
+- [Grating1DCalibration][extra.applications.Grating1DCalibration] can use the calibration mask
+  and use robust tools to improve the fit quality (!510).
```
tmichela, line 78: as I understand the code it adds masking, but also remove the background correction, right?

src/extra/applications/grating.py
```diff
@@ -275,18 +302,27 @@ def apply(self, run: DataCollection, load_all: bool=True) -> xr.Dataset:
         if load_all:
             out_data = run[self.grating_source, self.grating_key].xarray()
             trainId = out_data.trainId.to_numpy()
-            out_data = out_data.to_numpy() - self.bkg
+            out_data = out_data.to_numpy()
             out_data = out_data[:, self.offset::pulse_period, self.min_pixel:self.max_pixel]
+            if len(self.grating_mask_key) > 0:
```
tmichela: this is a string, right? I was confused for a second, I would just do `if self.grating_mask_key:` as the default is set to ""

src/extra/applications/grating.py
```diff
@@ -275,18 +302,27 @@ def apply(self, run: DataCollection, load_all: bool=True) -> xr.Dataset:
...
             out_data = list()
             for i, (tid, data) in enumerate(run.trains()):
-                #print(f"Train {tid}, idx {i}")
                 d = data[self.grating_source][self.grating_key]
-                if self.bkg is not None:
-                    d = d - self.bkg
                 # skip offset and collect pulse data each pulse_period samples only
                 d = d[self.offset::pulse_period, self.min_pixel:self.max_pixel]
+
+                if len(self.grating_mask_key) > 0:
```
tmichela: same here

tmichela (commented):
I think it LGTM. only few comments you can ignore.

Just one extra question: what happens if you do cls.from_file() from an old file that does not contain the new `grating_mask_key` parameter? I assume you can get an Attribute error when calling `apply()`.
Is this something that can happend?

src/extra/applications/grating.py
daniloefl: At the moment, I am the only one to create it, but I might not be the only one to `apply` it. The reason for the change is that this includes the full detector for which it was developed, so by default it should include all pixels and the user can restrict it. Actually, the full detector should go until 1280, so I will change it further.

docs/changelog.md
daniloefl: Right. I have changed it.

src/extra/applications/grating.py
daniloefl: Taken.

src/extra/applications/grating.py
daniloefl: Agreed.

daniloefl:
Thank you! Indeed your general point is a good one, so in the loading from file, I set the default for the `grating_mask` to the empty string, so that if it is not in the file, at least a default is set.

tmichela:
LGTM!

PR #511, opened 2026-07-01
TOFAnalogResponse: use thresholding to obtain a background-free response function

daniloefl:
TOFAnalogResponse aligns data from different conditions and averages them to extract the response function of a time-of-flight spectrometer. It can also simply count photo-electrons to obtain the detector response effect, ignoring the spectrometer impact. The latter mode is not very useful, as it disregards part of the response.

In this PR, the method of counting photo-electrons is used to threshold the data, while the analog data collected is still averaged out for the data passing the threshold. This allows this application to operate in very high-noise regimes.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -116,7 +116,7 @@ Changed:
 - [CookieboxCalibration][extra.applications.CookieboxCalibration] and
   [TOFAnalogResponse][extra.applications.TOFAnalogResponse] delegate
   parallelization to the `AdqRawChannel` class (!509).
-- [TOFAnalogResponse][extra.applications.TOFAnalogResponse] average analog pulses using thresholding
+- [TOFAnalogResponse][extra.applications.TOFAnalogResponse] averages analog pulses using thresholding
   from `AdqRawChannel` to clean up the data (!511).
 
 ## [2025.1]
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -116,6 +116,8 @@ Changed:
 - [CookieboxCalibration][extra.applications.CookieboxCalibration] and
   [TOFAnalogResponse][extra.applications.TOFAnalogResponse] delegate
   parallelization to the `AdqRawChannel` class (!509).
+- [TOFAnalogResponse][extra.applications.TOFAnalogResponse] average analog pulses using thresholding
+  from `AdqRawChannel` to clean up the data (!511).
 
 ## [2025.1]
 
--- a/src/extra/applications/cookiebox_deconvolve.py
+++ b/src/extra/applications/cookiebox_deconvolve.py
@@ -487,13 +487,13 @@ class TOFAnalogResponse(SerializableMixin):
                 logging.info("Calling pulse_edges ...")
                 tof_data = tof.pulse_edges(pulse_dim='pulseIndex', threshold=self.count_threshold, parallel=parallel).reset_index()
                 good_trains = np.unique(tof_data.loc[:,'trainId'].to_numpy())
-                idx = tof_data.loc[:, ['trainId', 'pulseIndex']].set_index(['trainId', 'pulseIndex']).idx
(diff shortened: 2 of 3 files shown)


src/extra/applications/cookiebox_deconvolve.py
```diff
@@ -470,19 +470,27 @@ def setup(self,
             else:
                 # count photo-electrons by histogramming peak positions
                 for k, e in enumerate(scan.positions):
+                    logging.info("Calling pulse_edges for selected trains ...")
                     tof_data = tof.select_trains(by_id[scan.positions_train_ids[k]]).pulse_edges(pulse_dim='pulseIndex', threshold=self.count_threshold, parallel=parallel).reset_index()
```
takluyver: We can simplify this a little bit - because the ADQ component has `.select_trains()`, we can split it with the scan:


```suggestion
                for tof_part in scan.split_by_steps(tof):
                    logging.info("Calling pulse_edges for selected trains ...")
                    tof_data = tof_part.pulse_edges(pulse_dim='pulseIndex', threshold=self.count_threshold, parallel=parallel).reset_index()
```

We could probably even go a step further, getting rid of the loop, loading the pulse data for all scan steps, and then using `scan.group_data(analog_data.sel(pulse=idx)).mean()` to make an average for each scan step. But that's a slightly bigger change.

src/extra/applications/cookiebox_deconvolve.py
```diff
@@ -470,19 +470,27 @@ def setup(self,
...
             else:
+                logging.info("Calling pulse_edges ...")
                 tof_data = tof.pulse_edges(pulse_dim='pulseIndex', threshold=self.count_threshold, parallel=parallel).reset_index()
-                this_tof_data, _ = np.histogram(tof_data.edge, bins=bins, weights=-tof_data.amplitude)
-                this_tof_data = xr.DataArray(this_tof_data, dims=('sample'), coords={'sample': bins[:-1]})
+                good_trains = np.unique(tof_data.loc[:,'trainId'].to_numpy())
+                idx = tof_data.loc[:, ['trainId', 'pulseIndex']].set_index(['trainId', 'pulseIndex']).index
+                logging.info("Summing good trains ...")
+                analog_data = -tof.select_trains(by_id[good_trains]).pulse_data(pulse_dim="pulseIndex", parallel=parallel)
+                this_tof_data = analog_data.sel(pulse=idx).mean("pulse")
```
takluyver: This is pretty similar to the code above. If we do keep the loop for the scan case, I think there might be enough complexity to be worth deduplicating them as a little helper function. It's admittedly a borderline case, though.

docs/changelog.md
```diff
@@ -116,6 +116,8 @@ Changed:
 - [CookieboxCalibration][extra.applications.CookieboxCalibration] and
   [TOFAnalogResponse][extra.applications.TOFAnalogResponse] delegate
   parallelization to the `AdqRawChannel` class (!509).
+- [TOFAnalogResponse][extra.applications.TOFAnalogResponse] average analog pulses using thresholding
```
takluyver:
```suggestion
- [TOFAnalogResponse][extra.applications.TOFAnalogResponse] averages analog pulses using thresholding
```

takluyver (commented):
LGTM, all suggestions are minor.

src/extra/applications/cookiebox_deconvolve.py
daniloefl: Cool! I didn't know about this. I took your suggestion.

src/extra/applications/cookiebox_deconvolve.py
daniloefl: Indeed, I did some re-organization to move this to a separate function. Thank you!

docs/changelog.md
daniloefl: Of course!

takluyver:
Thanks, LGTM

daniloefl:
Thank you, Thomas!

PR #512, opened 2026-07-01
CookieboxCalibration: Normalize transmission by pulse energy and detect RoI robustly

daniloefl:
This PR contains two elements:
1. The transmission calculation is affected currently due to variations in the pulse energy during the data collection. This is corrected here by normalizing not only by the angular dependence given by the FEL polarization, but also by the pulse energy within the data collected (which is now also calculated and stored for this purpose).
2. The region of interest containing the photoline can be given or auto-detected if it can be assumed the only other peak comes from the Auger-Meitner. In this PR, the autodetection is slightly improved by smoothing the signal beforehand.

### DIFF
--- a/src/extra/applications/cookiebox.py
+++ b/src/extra/applications/cookiebox.py
@@ -58,11 +58,20 @@ def search_roi(roi: np.ndarray) -> Tuple[np.ndarray, np.ndarray]:
     Returns: Peak position.
     """
     roi_smooth = gaussian_filter(roi, sigma=5)
-    p, prop = find_peaks(roi_smooth, prominence=0.25*np.max(roi), width=1)
-    idx = np.argsort(p)[::-1]
+    p, prop = find_peaks(roi_smooth, prominence=0.25*np.max(roi_smooth), width=1, height=0)
+    idx = np.argsort(prop["peak_heights"])[::-1]
     p = p[idx]
     w = prop["widths"][idx]
-    return p, w
+    if len(p) == 0:
+        logging.info(f"Failed to find peaks.")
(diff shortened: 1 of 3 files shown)


src/extra/applications/cookiebox.py
```diff
@@ -56,9 +57,21 @@ def search_roi(roi: np.ndarray) -> np.ndarray:
 
     Returns: Peak position.
     """
-    import scipy
-    p, _ = scipy.signal.find_peaks(roi, prominence=(0.25*np.max(roi), None))
-    return p
+    roi_smooth = gaussian_filter(roi, sigma=5)
```
JamesWrigley: Nitpicking, will the input ever contain nonfinite values like nans that should be ignored?

JamesWrigley (commented):
LOCKED IN 🚀

src/extra/applications/cookiebox.py
daniloefl: That is definitely not nitpicking. It should not contain `nan` or `inf`. But what would be your proposal if it did? If there is a simple fix in this case, I am open to change it.

JamesWrigley: I would do something like `gaussian_filter(roi[np.isfinite(roi)])`. But if the input shouldn't contain it then no worries.

JamesWrigley (approved):
LGTM!

(but pls squash merge)

src/extra/applications/cookiebox.py
daniloefl: I think it is a good suggestion. I took it anyway.

PR #513, opened 2026-07-01
CookieboxCalibration: added flag to parallelize over etofs

daniloefl:
Use `ProcessPoolExecutor` to parallelize over eTOFs. There are two parallelization flags and the default (tested to be optimal) is to not parallelize over the IO (ie, no parallelization on `AdqRawChannel`), but to parallelize over the different eTOFs.


philsmt:
LGTM

PR #515, opened 2026-07-15
Include motor movements in the fit.

daniloefl:
Adds motor positions given to the fit, so that motor positions are compensated in the calibration.

### DIFF
--- a/src/extra/applications/grating.py
+++ b/src/extra/applications/grating.py
@@ -74,13 +74,17 @@ class Grating1DCalibration(SerializableMixin):
                             "max_pixel",
                             "e0",
                             "slope",
+                            "slope_motor",
                             "energy_axis",
                             "calibration_energies",
                             "calibration_data",
+                            "calibration_motor",
                             "grating_source",
                             "grating_key",
                             "sources",
                             "grating_mask_key",
+                            "grating_motor_source",
+                            "grating_motor_key",
                             "sigma",
                             "_version",
                            ]
@@ -90,6 +94,7 @@ class Grating1DCalibration(SerializableMixin):
               scan: Scan,
               pulses: XrayPulses,
               grating_mask: Optional[KeyData]=None,
+              grating_motor: Optional[KeyData]=None,
               ):
         """
         Setup calibration.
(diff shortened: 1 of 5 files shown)


src/extra/applications/grating.py
```diff
@@ -232,10 +250,18 @@ def load_data(self):
...
+                     grating=self._grating_motor,
+                     )
         energy_ids = np.arange(len(self.calibration_energies))
         # average data in each mono scan bin
         with ProcessPoolExecutor() as p:
             data = np.stack(list(p.map(fn, energy_ids)), axis=0)
+            if self.grating_motor_source != "":
+                data_motor = np.stack(list(p.map(fn_motor, energy_ids)), axis=0)
+            else:
+                data_motor = np.zeros((data.shape[0]))
```
JamesWrigley: Nitpicking:
```suggestion
                data_motor = np.zeros(data.shape[0])
```

src/extra/applications/grating.py
```diff
@@ -261,16 +288,23 @@ def mask_calibration_point(self, energy: float, mask: bool=False, tol: float=1.0
...
+        #from scipy.stats import linregress
         mask = self.calibration_mask
         sample = np.arange(self.calibration_data.shape[-1])
         sample_mode = np.nanargmax(self.calibration_data, axis=-1)
-        x = sample_mode[mask]
+        motor_position = self.calibration_motor
+        #sample_mode = np.sum(self.calibration_data*sample, axis=-1)/np.sum(self.calibration_data, axis=-1)
+        #res = linregress(sample_mode[mask], self.calibration_energies[mask])
+        #self.slope = res.slope
+        #self.e0 = res.intercept
```
JamesWrigley: Could these be deleted?

src/extra/applications/grating.py
```diff
@@ -280,24 +314,42 @@ def plot(self):
...
-    def apply(self, run: DataCollection, load_all: bool=True) -> xr.Dataset:
+    def apply(self, run: DataCollection, load_all: bool=True, assume_motor=None) -> xr.Dataset:
         """
         Apply calibration to a new analysis run.
         It is assumed it contains the same settings.
 
         Args:
           run: Input run.
-          load_all: If True, load all data in memory at once. This is faster, but uses more memory.
+          load_all: If True, load all data in memory at once. This is faste, but uses more memory.
```
JamesWrigley: Nitpicking:
```suggestion
          load_all: If True, load all data in memory at once. This is faster, but uses more memory.
```

JamesWrigley (changes requested):

src/extra/applications/grating.py
daniloefl: Good point.

src/extra/applications/grating.py
daniloefl: Yes. Good point. I did it.

JamesWrigley (approved):
LGTM!

(but pls squash merge 😅)

PR #519, opened 2026-08-03
Update BadPixels with new bits

philsmt:
Some updates to the `BadPixels` enum as discussed in CAL dev team:

* `EXPERT_CHOICE` for interactive masks set by the operators and experts, e.g. the `BadPixelsManual` 
* Reserved last two bits for arbitrary use labelled `USER1` and `USER2`


philsmt:
@takluyver will add the changelog entry directly to https://github.com/European-XFEL/EXtra/pull/518

PR #538, opened 2026-08-25
Improve numerical stability of the deconvolution in TOFAnalogDeconvolution

daniloefl:
As part of the deconvolution algorithm, it is necessary to obtain the inverse of a matrix $I + \lambda A^T A$, where $I$ is the identity, $\lambda \in \mathbb{R}$, and $A$ is a square real matrix. So far `pinv` has been used to do that, but this matrix is self-adjoint, so one could improve speed and numerical stability by using `pinvh`, which assumes the upper-triangular part of the matrix equals the lower triangular part (in case of a real matrix).

Additionally, in case of numerical instability when extracting the response function, since there are multiple estimates of the response function that are averaged, if the deconvolution fails in one case, this sample is ignored and the remaining ones are used. This is done by catching a `LinAlgError` when doing the deconvolution.

### DIFF
--- a/src/extra/applications/cookiebox_deconvolve.py
+++ b/src/extra/applications/cookiebox_deconvolve.py
@@ -1,6 +1,6 @@
 from typing import Tuple, Union, Optional, List
 from functools import partial
-from scipy.linalg import convolution_matrix
+from scipy.linalg import convolution_matrix, pinvh
 from scipy.signal import find_peaks
 import numpy as np
 import xarray as xr
@@ -79,7 +79,7 @@ def nn_deconvolution(data: np.ndarray, h: np.ndarray, n_iter: int=4000, n_shift:
     I = np.eye(N)
 
     rho = 0.1
-    G = np.linalg.pinv(A.T @ A + rho*I)
+    G = pinvh(A.T @ A + rho*I)
     x = np.copy(b)
     z = np.copy(b)
     u = np.copy(b)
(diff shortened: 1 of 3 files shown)


src/extra/applications/cookiebox_deconvolve.py
```diff
@@ -195,7 +195,7 @@ def tv_deconvolution(data: np.ndarray, h: np.ndarray, Lambda: float=1e-5, n_iter
     # Its proximal operator is tabulated (see Ref. [2], sec. 6.1.1):
     # $\text{prox}_{\tau G} (v) = (I + \tau A^T A)^{-1} (v + \tau A^T b)
     I = np.eye(N)
-    GG = np.linalg.pinv(I + tau*(A.T @ A))
+    GG = pinvh(I + tau*(A.T @ A))
```
takluyver: I see the [numpy pinv function](https://numpy.org/doc/stable/reference/generated/numpy.linalg.pinv.html#numpy.linalg.pinv) can be passed `hermitian=True` to make the same assumption - any idea how this compares to the scipy function?

(I'm not saying the numpy one is better - I have no idea)

daniloefl: I also do not fully understand, but based on [this](https://stackoverflow.com/questions/13265299/the-difference-of-pseudo-inverse-between-scipy-and-numpy), they solve different mathematical problems (see [1] and [2] for the underlying Fortran code documentation). In one case the divide-and-conquer algorithm for SVD is used (in numpy), in the other a least squares problem is solved.

It seems that the numpy implementation (divide-and-conquer) is faster and much more memory efficient, but it is unclear if the accuracy is better. I have tried to switch to the numpy strategy and it seems to work as well and no numerical instabilities problems occurred in situations where I had observed it before.

As a result, I have switched indeed to the numpy implementation.

Refs.:
[1] https://netlib.org/lapack/explore-html/df/d22/group__gesdd_ga8941e5ff50de36580dae8940015e9cb0.html
[2] https://netlib.org/lapack/explore-html//da/d55/group__gelss_gac6159de3953ae0386c2799294745ac90.html

takluyver:
LGTM

PR #547, opened 2026-09-09
Add support for `juliacall.JlArray` in `imshow2()`

JamesWrigley:
The type was renamed in an upcoming PythonCall release.


JamesWrigley:
Sneaked in another fix for the scantool component in 94923e3.

JamesWrigley:
(merging because it's annoying to work around)

